PR #1, opened 2016-02-24
JWT key is now accessible only on the Liana

SeyZ:


SeyZ:
@arnaudbesnier let's discuss about this

PR #2, opened 2016-06-06
Create/Update - Fix the create and update actions on Rails 4

arnaudbesnier:


lib/forest_liana/version.rb
```diff
@@ -1,3 +1,3 @@
 module ForestLiana
-  VERSION = "1.3.3"
+  VERSION = "1.3.6"
```
SeyZ: @arnaudbesnier I prefer to change the version manually on each release

PR #6, opened 2016-06-07
Charts - Order Pie charts by count/sum

arnaudbesnier:


app/services/forest_liana/pie_stat_getter.rb
```diff
@@ -17,8 +17,14 @@ def perform
                                                 filter_value)
         end
 
+        field = 'all'
```
SeyZ: Should be awesome to have a comment to explain why 'all' here :)

PR #14, opened 2016-07-25
Filters - Users wants to have filters that work on "class_name" belongs_to

arnaudbesnier:


app/services/forest_liana/search_query_builder.rb
```diff
@@ -51,7 +51,7 @@ def search_param
     def filter_param
       if @params[:filter]
         @params[:filter].each do |field, values|
-          next if has_many_association?(field)
+          next if association?(field)
```
arnaudbesnier: Double filter because belongs-to associations where not stopped here

app/services/forest_liana/search_query_builder.rb
```diff
@@ -141,12 +141,18 @@ def belongs_to_subfield_filter(field, value)
...
+
+      operator_date_interval_parser = OperatorDateIntervalParser.new(value)
+      if operator_date_interval_parser.is_interval_date_value()
+        filter = operator_date_interval_parser.get_interval_date_filter()
+        @records = @records.where("#{association.table_name}.#{subfield} #{filter}")
+      else
+        where = "#{association.table_name}.#{subfield} #{operator}"
+        where += " '#{value}'" if value
+        @records = @records.where(where)
+      end
```
arnaudbesnier: Fix filters date operator filters on belongs_to association (due to the fix)

PR #18, opened 2016-08-01
Apimap - Fix duplicate collections in the apimap

arnaudbesnier:


lib/forest_liana/bootstraper.rb
```diff
@@ -50,7 +50,7 @@ def create_apimap
       SchemaUtils.tables_names.map do |table_name|
         model = SchemaUtils.find_model_from_table_name(table_name)
         if model.try(:table_exists?)
-          ForestLiana.apimap << SchemaAdapter.new(model).perform
+          SchemaAdapter.new(model).perform
```
arnaudbesnier: Not sure if it is the right fix to apply.
Anyway, each collection was pushed twice and this change removes one push.

PR #20, opened 2016-08-03
Activity Log - Users want to see when an action is triggered on the activity log

arnaudbesnier:


app/controllers/forest_liana/actions_controller.rb
```diff
@@ -0,0 +1,14 @@
+module ForestLiana
+  class ActionsController < ForestLiana::ApplicationController
```
arnaudbesnier: This change needs to change custom actions implémentation on the client app side (see https://github.com/ForestAdmin/forest-movies/pull/2/files). Thoughts?

PR #32, opened 2016-09-20
Charts - Fix value chart with filters and a previous period comparison

arnaudbesnier:


app/services/forest_liana/operator_value_parser.rb
```diff
@@ -3,30 +3,32 @@ class OperatorValueParser
 
     def self.parse(value)
```
arnaudbesnier: The `parse` method was modifying the value `passed` as a parameter. If we pass at least twice in the method, the query generated is bad.
That was the case for a value chart with filters and a previous period comparison: 2 requests are done on a different time range.

PR #40, opened 2016-10-10
Polymorphism - Fix the nullification of a polymorphic has_one association

arnaudbesnier:


app/services/forest_liana/belongs_to_updater.rb
```diff
@@ -9,13 +9,8 @@ def initialize(resource, association, params)
...
       @record = @resource.find(@params[:id])
-
-      if @data && @data[:id]
-        new_value = @association.klass.find(@data[:id]) if @data && @data[:id]
-        @record.send("#{@association.name}=", new_value)
-      else
-        @record.send("#{@association.foreign_key}=", nil)
-      end
+      new_value = @association.klass.find(@data[:id]) if @data && @data[:id]
+      @record.send("#{@association.name}=", new_value)
```
arnaudbesnier: it also makes the code cleaner than I did it in the first version...

PR #43, opened 2016-10-12
Users can search on Smart fields

SeyZ:


app/services/forest_liana/resources_getter.rb
```diff
@@ -16,7 +16,7 @@ def records
     end
 
     def count
-      search_query.count
+      @records.count
```
arnaudbesnier: This change is strange, is it really wanted?

SeyZ: Yes it is. We need to have the last updated @records (not just the search_query).

PR #45, opened 2016-10-13
Users see active record errors

SeyZ:


app/controllers/forest_liana/resources_controller.rb
```diff
@@ -29,16 +29,26 @@ def create
       creator = ResourceCreator.new(@resource, params)
       creator.perform
 
-      render serializer: nil,
-        json: serialize_model(creator.record, include: includes)
+      if creator.record.invalid?
```
arnaudbesnier: I think that the `valid?` method would make the code more readable. The first block would be the `success` case and the second the `error` case.

PR #49, opened 2016-10-18
Users want to exclude models

SeyZ:


lib/forest_liana.rb
```diff
@@ -13,11 +13,13 @@ module UserSpace
   mattr_accessor :apimap
   mattr_accessor :allowed_users
   mattr_accessor :models
+  mattr_accessor :exclude_models
```
arnaudbesnier: `models_excluded` ?

lib/forest_liana/bootstraper.rb
```diff
@@ -84,7 +92,7 @@ def check_integrations_setup
         else
           @logger.error 'Cannot setup properly your Stripe integration.'
         end
-      end
+     end
```
arnaudbesnier: The indentation was correct here, but broken at the `if` level I guess.

lib/forest_liana.rb
arnaudbesnier: `models_excluded` would really be more consistant with `models` variable

PR #54, opened 2016-10-21
Support STI models

SeyZ:


lib/forest_liana/bootstraper.rb
```diff
@@ -45,13 +45,26 @@ def perform
 
     private
 
+    def analyze_sti_models(model)
```
arnaudbesnier: fetch_sti_models would be more explicit I think

PR #57, opened 2016-10-27
Rails 5 users want to see their models

SeyZ:


lib/forest_liana/bootstraper.rb
```diff
@@ -45,7 +45,7 @@ def perform
 
     private
 
-    def analyze_sti_models(model)
+    def fetch_sti_models(model)
```
arnaudbesnier: :P

PR #62, opened 2016-11-05
Polymorphism - Fix polymorphic "hasMany" on record hasMany association retrieval

arnaudbesnier:


app/services/forest_liana/has_many_getter.rb
```diff
@@ -31,9 +31,9 @@ def includes
       @association.klass
         .reflect_on_all_associations
         .select do |association|
-          inclusion = SchemaUtils.model_included?(association.klass) &&
-            [:belongs_to, :has_and_belongs_to_many].include?(association.macro) &&
-            !association.options[:polymorphic]
+          inclusion = !association.options[:polymorphic] &&
```
arnaudbesnier: Just move the `!association.options[:polymorphic]` condition to the first position. As it is a lazy evaluation, the polymorphic associations won't be evaluated.

PR #72, opened 2016-11-17
Made user_class_name configurable

GabKlein:
Hi guys,

Here a quick update that allows us to configure the `ForestUser`.

### DIFF
--- a/app/controllers/forest_liana/sessions_controller.rb
+++ b/app/controllers/forest_liana/sessions_controller.rb
@@ -21,11 +21,11 @@ module ForestLiana
 
     def check_user
       # NOTICE: Use the ForestUser table for authentication.
-      if defined? ForestUser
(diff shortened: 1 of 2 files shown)


GabKlein:
@SeyZ Any update on this one? Let me know what would fit the best your implementation.

> Note: if we want to be rails agnostic we can use something in the same vein as:
```ruby
Object.const_defined?(ForestLiana.user_class_name) && Object.const_get(ForestLiana.user_class_name)
```
or 
```ruby
Object.const_get(ForestLiana.user_class_name) rescue nil
```

arnaudbesnier:
Hey @GabKlein, the current implementation breaks the "classical" user authentication usage. I have upgraded your commit with the fix here: https://github.com/ForestAdmin/forest-rails/pull/75.

Let me know if it is fine for you.

PR #74, opened 2016-11-18
Pie Charts - Support Pie charts with a group by on an association field

arnaudbesnier:


SeyZ:
Just tested on Movies with a Pie chart:
Comment - Count - groupBy movie:

```ActiveRecord::StatementInvalid (PG::UndefinedColumn: ERROR:  column comments.movie does not exist
LINE 1: ...LECT COUNT(DISTINCT "comments"."id") AS count_id, comments.m...
                                                             ^
: SELECT COUNT(DISTINCT "comments"."id") AS count_id, comments.movie AS comments_movie FROM "comments" LEFT OUTER JOIN "customers" ON "customers"."id" = "comments"."cust
omer_id" LEFT OUTER JOIN "movies" ON "movies"."id" = "comments"."movie_id" GROUP BY comments.movie  ORDER BY count_id DESC):```

PR #77, opened 2016-11-21
Allow to override the read_only option of a Smart Field

SeyZ:


lib/forest_liana/collection.rb
```diff
@@ -14,16 +14,19 @@ def collection(collection_name, opts = {})
...
     def field(name, opts, &block)
+      opts[:read_only] = true unless opts.has_key?(:read_only)
+      opts[:is_searchable] = false unless opts.has_key?(:is_searchable)
+
       model.fields << opts.merge({
         field: name,
-        :'is-read-only' => true,
-        :'is-searchable' => false,
+        :'is-read-only' => opts[:read_only],
+        :'is-searchable' => opts['is_searchable'],
```
arnaudbesnier: Will it activate the filters on such a smart field if `is_searchable` is set to `true`?

lib/forest_liana/collection.rb
```diff
@@ -14,16 +14,19 @@ def collection(collection_name, opts = {})
...
 
     def field(name, opts, &block)
+      opts[:read_only] = true unless opts.has_key?(:read_only)
+      opts[:is_searchable] = false unless opts.has_key?(:is_searchable)
+
       model.fields << opts.merge({
         field: name,
-        :'is-read-only' => true,
-        :'is-searchable' => false,
+        :'is-read-only' => opts[:read_only],
```
arnaudbesnier: I guess you have to implement a `setter` method to activate the write mode. The API should set the property to `false` automatically if the setter is defined. What do you think?

SeyZ: Not sure to understand, here's how to set the readonly option on a smart field

```
field :fullname, type: 'String', read_only: false do
  # ...
end
```

Is it ok?

lib/forest_liana/collection.rb
SeyZ: yes

PR #80, opened 2016-11-23
Add resource table to avoid ambigous column name error

tsara27:
I encountered this error when loading my table.
```
ActiveRecord::StatementInvalid (PG::AmbiguousColumn: ERROR:  column reference "id" is ambiguous
LINE 1: ..." AS t7_r26, "sessions"."coachee_type" AS t7_r27, id, client...
```
The error will be raised when calling `getter.records` on `resources_controller.rb`. So, I modified the code a bit to fix that. Thanks

### DIFF
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -110,7 +110,9 @@ module ForestLiana
     end
 
     def select
-      column_names = @resource.column_names.map { |name| name.to_sym }
+      column_names = @resource.column_names.map do |name|
+        "#{@resource.table_name}.#{name}".to_sym
+      end
       if @field_names_requested
         column_names & @field_names_requested
       else


arnaudbesnier:
Hey @tsara27,

Thanks for the contribution!
We'll definitely review the pull request this week.

app/services/forest_liana/resources_getter.rb
```diff
@@ -110,7 +110,9 @@ def association?(field)
     end
 
     def select
-      column_names = @resource.column_names.map { |name| name.to_sym }
+      column_names = @resource.column_names.map do |name|
+        "#{@resource.table_name}.#{name}".to_sym
```
arnaudbesnier: Hey @tsara27, the change you made is obviously going to the right direction but it broke the performance optimizations we implemented below. We only want to select the column names found in the request params.

arnaudbesnier:
Thanks for your contribution @tsara27, I have improved your fix in this new PR: https://github.com/ForestAdmin/forest-rails/pull/81

🌳

tsara27:
Thank you @arnaudbesnier

PR #83, opened 2016-11-29
Date Filters - Date filters operators are now based on the client timezone

arnaudbesnier:


app/services/forest_liana/line_stat_getter.rb
```diff
@@ -16,34 +20,17 @@ def perform
...
       value = value.send(@params[:aggregate].downcase, @params[:aggregate_field])
         .map do |k, v|
-          if k.kind_of?(Array)
-            {
-              label: k[0],
-              values: {
-                key: populate(k[1]),
-                value: v
-              }
-            }
```
arnaudbesnier: This is useless and was creating buggy line charts

app/services/forest_liana/line_stat_getter.rb
```diff
@@ -55,30 +42,6 @@ def group_by_date_field
...
-        field_name = @params[:group_by_field]
-        association = @resource.reflect_on_association(field_name)
-
-        if association
-          association.klass.find(id)
-        else
-          id
-        end
-      end
-    end
```
arnaudbesnier: Those 2 fonctions are not relevant

PR #87, opened 2016-12-08
Search - Support the search for MySQL databases

arnaudbesnier:


app/helpers/forest_liana/adapter_helper.rb
```diff
@@ -0,0 +1,8 @@
+module ForestLiana
+  module AdapterHelper
+    def self.format_column_name(name)
+      separator = ActiveRecord::Base.connection.adapter_name == "MySQL" ? '`' : '"'
+      "#{separator}#{name.gsub('.', "#{separator}.#{separator}")}#{separator}"
```
SeyZ: @arnaudbesnier Why don't you use `ActiveRecord::Base.connection.quote_column_name "my_column"` ?
http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Quoting.html#method-i-quote_column_name

PR #107, opened 2017-02-24
Test uploaded file against Base64 regexp for Carrierwave

Borzik:
When I update a model via Forest Admin, browser submits image attribute even if it's marked readonly and if it hasn't changed. In this case browser passes a URL for the existing image, which caused `ForestLiana::Base64StringIO.initialize` to throw an `ArgumentError`.

This PR is to ignore carrierwave fields (by removing related key from attributes hash) which contain incorrect (non-base64 data) strings.


arnaudbesnier:
Hi @Borzik, thanks for the contribution, we'll look into your code soon.

PR #111, opened 2017-03-14
[-] CORS - Enable all forestadmin.com subdomains to requests the liana

arnaudbesnier:
Must have to make Rails users test using the new designed frontapp.

### DIFF
--- a/lib/forest_liana/engine.rb
+++ b/lib/forest_liana/engine.rb
@@ -19,8 +19,7 @@ module ForestLiana
 
         config.middleware.insert_before 0, rack_cors_class do
           allow do
-            hostnames = ['localhost:4200', 'app.forestadmin.com',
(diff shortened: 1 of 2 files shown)


lib/forest_liana/engine.rb
```diff
@@ -19,8 +19,7 @@ def configure_forest_cors
 
         config.middleware.insert_before 0, rack_cors_class do
           allow do
-            hostnames = ['localhost:4200', 'app.forestadmin.com',
-                         'www.forestadmin.com']
+            hostnames = ['localhost:4200', '*.forestadmin.com']
```
SeyZ: Not sure the wildcards function using rack-cors.
I think you have to pass a Regexp.

By the way, our customers can use the CORS_ORIGINS env variable to add v2.forestadmin.com

PR #126, opened 2017-04-04
Fix the eager_load COUNT

SeyZ:


arnaudbesnier:
Fixed in the front app

PR #128, opened 2017-04-05
Avoid to analyze model with the same table_names (STI)

SeyZ:


SeyZ:
[Users don't want to see duplicates fields](https://trello.com/c/f0fWqiNx/1371-users-don-t-want-to-see-duplicates-fields)

SeyZ:
@arnaudbesnier I've reproduced the issue and the problem is related to warnings in the console when using STI models.

lib/forest_liana/bootstraper.rb
```diff
@@ -32,8 +32,11 @@ def perform
     private
 
     def fetch_sti_models(model)
-      type_field = model.columns.find { |c| c.name == 'type' }
-      if type_field
+      inheritance_field = model.columns.find {
+        |c| c.name == model.inheritance_column || c.name == 'type'
+      }
```
SeyZ: Rails allows the user to override the default `type` field to specify this model is a STI one.

lib/forest_liana/bootstraper.rb
```diff
@@ -44,7 +47,7 @@ def fetch_sti_models(model)
 
     def analyze_model?(model)
       return model && model.table_exists? && !SchemaUtils.habtm?(model) &&
-        SchemaUtils.model_included?(model)
+        SchemaUtils.model_included?(model) && !model.parent.try(:table_name)
```
SeyZ: If the current model has a parent and this parent has a table_name, it means the model is already present in Forest. Do not add it twice (it's the case for STI, but it could also be the case for other use cases).

lib/forest_liana/bootstraper.rb
arnaudbesnier: Not really confident with the last part `|| c.name == 'type'` but I don't really know why 😄 . Is `inheritance_column` always defined (default or overriden field)?

arnaudbesnier: It needs a dedicated change in the changelog to explain that the liana support overriden STI type fields

lib/forest_liana/bootstraper.rb
arnaudbesnier: Does it mean that STI models are not sent in the apimap?

lib/forest_liana/bootstraper.rb
SeyZ: Agree for the CHANGELOG.
Because by default, Rails consider a model as a STI if it contains a field name 'type', nothing more :)

SeyZ:
@arnaudbesnier updated, can you review one more time plz?

app/services/forest_liana/schema_adapter.rb
```diff
@@ -270,6 +271,23 @@ def add_enum_values_if_is_enum(column_schema, column)
...
+      end
+
+      column_schema
+    end
+
+    def sti_column?(column)
+      (@model.inheritance_column &&
+       column.name == @model.inheritance_column) || column.name == 'type'
+    end
+
```
arnaudbesnier: 👍

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
```
arnaudbesnier: ### Changed
- STI Models - Add a select input with predefined existing STI type values in create/update forms.
- STI Models - Children models do not generate a specific collection anymore (only parent model). 

(something like this)

PR #130, opened 2017-04-06
Ensure to add the parent STI models to the ForestLiana.models

SeyZ:


lib/forest_liana/bootstraper.rb
```diff
@@ -33,7 +33,7 @@ def perform
 
     def analyze_model?(model)
       return model && model.table_exists? && !SchemaUtils.habtm?(model) &&
-        SchemaUtils.model_included?(model) && !SchemaUtils.sti?(model)
+        SchemaUtils.model_included?(model) && !SchemaUtils.sti_child?(model)
```
SeyZ: We want to keep the parent, but not the children

Issue #132, opened 2017-04-10
Forest prevents app/console from starting when disconnected from internet


josephpage:
I was disconnected from internet, I just tried to launch `rails console` and 💥 : 

```
/usr/local/lib/ruby/2.4.0/net/http.rb:906:in `rescue in block in connect': Failed to open TCP connection to forestadmin-server.herokuapp.com:443 (getaddrinfo: Name or service not known) (SocketError)
	from /usr/local/lib/ruby/2.4.0/net/http.rb:903:in `block in connect'
	from /usr/local/lib/ruby/2.4.0/timeout.rb:93:in `block in timeout'
	from /usr/local/lib/ruby/2.4.0/timeout.rb:103:in `timeout'
	from /usr/local/lib/ruby/2.4.0/net/http.rb:902:in `connect'
	from /usr/local/lib/ruby/2.4.0/net/http.rb:887:in `do_start'
	from /usr/local/lib/ruby/2.4.0/net/http.rb:876:in `start'
	from /usr/local/bundle/gems/forest_liana-1.5.23/lib/forest_liana/bootstraper.rb:146:in `send_apimap'
	from /usr/local/bundle/gems/forest_liana-1.5.23/lib/forest_liana/bootstraper.rb:28:in `perform'
	from /usr/local/bundle/gems/forest_liana-1.5.23/lib/forest_liana/engine.rb:51:in `block in <class:Engine>'
	from /usr/local/bundle/gems/activesupport-5.1.0.rc1/lib/active_support/lazy_load_hooks.rb:43:in `execute_hook'
	from /usr/local/bundle/gems/activesupport-5.1.0.rc1/lib/active_support/lazy_load_hooks.rb:52:in `block in run_load_hooks'
	from /usr/local/bundle/gems/activesupport-5.1.0.rc1/lib/active_support/lazy_load_hooks.rb:51:in `each'
	from /usr/local/bundle/gems/activesupport-5.1.0.rc1/lib/active_support/lazy_load_hooks.rb:51:in `run_load_hooks'
	from /usr/local/bundle/gems/railties-5.1.0.rc1/lib/rails/application/finisher.rb:73:in `block in <module:Finisher>'
	from /usr/local/bundle/gems/railties-5.1.0.rc1/lib/rails/initializable.rb:30:in `instance_exec'
	from /usr/local/bundle/gems/railties-5.1.0.rc1/lib/rails/initializable.rb:30:in `run'
[...]

PR #133, opened 2017-04-11
[*] Server Start - Fix the server crash on start while offline

arnaudbesnier:
Fixes: https://github.com/ForestAdmin/forest-rails/issues/132

### DIFF
--- a/lib/forest_liana/bootstraper.rb
+++ b/lib/forest_liana/bootstraper.rb
@@ -150,20 +150,25 @@ module ForestLiana
             request['forest-secret-key'] = ForestLiana.env_secret
             response = client.request(request)
 
-            unless response.body.blank?
-              warning = JSON.parse(response.body)['warning']
-            end
+            if ['200', '202', '204', '400', '404'].include? response.code
+              unless response.body.blank?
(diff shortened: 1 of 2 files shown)


lib/forest_liana/bootstraper.rb
```diff
@@ -150,23 +150,28 @@ def send_apimap
...
-              FOREST_LOGGER.error "Cannot find the project related to the " \
-                "env_secret you configured. Can you check on Forest that you " \
-                "copied it properly in the forest_liana initializer?"
-            elsif response.is_a?(Net::HTTPBadRequest) # NOTICE: HTTP 400 Error
-              FOREST_LOGGER.error "An error occured with the apimap sent to " \
-                "Forest. Please contact support@forestadmin.com for further " \
-                "investigations."
-            elsif warning
-              FOREST_LOGGER.warn warning
+            if ['200', '202', '204', '400', '404'].include? response.code
```
SeyZ: What happen if the error code is 500?

arnaudbesnier: Maybe you fall in the `else` case 😆

PR #135, opened 2017-04-19
[*] Filters ToDate - Fix the end of period filtering for "toDate" date operator types

arnaudbesnier:


app/services/forest_liana/operator_date_interval_parser.rb
```diff
@@ -83,7 +83,7 @@ def get_interval_date_filter
       if match && match[1]
         return "BETWEEN '" +
           "#{to_client_timezone((Integer(match[1]) - 1).day.ago.beginning_of_day)}'" +
-          " AND '#{to_client_timezone(Time.now)}'"
+          " AND '#{Time.now}'"
```
SeyZ: Can you explain why the end of the period does not take the timezone into account?

arnaudbesnier: because now is now (whatever is your timezone). Here the generated date is the current time.

PR #136, opened 2017-04-19
[*] Record Deletion - Fix the records deletion on Rails 5.1

arnaudbesnier:


app/controllers/forest_liana/resources_controller.rb
```diff
@@ -64,9 +64,7 @@ def update
     end
 
     def destroy
-      selector = {}
-      selector[@resource.primary_key] = params[:id]
-      @resource.destroy_all(selector)
+      @resource.destroy(params[:id])
```
SeyZ: Are you sure we don't want to keep `destroy_all`? Why?
(http://stackoverflow.com/questions/6698207/delete-all-vs-destroy-all)

SeyZ: Ok it's the same

PR #139, opened 2017-04-27
[+] Acts on taggable - Search & Filters

SeyZ:
Not sure about the implementation, would love to have your feedback @arnaudbesnier and @lucasscariot

### DIFF
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -60,6 +60,18 @@ module ForestLiana
           end
         end
 
+        # ActsAsTaggable
+        if @resource.respond_to?(:acts_as_taggable)
+          @resource.acts_as_taggable.each do |field|
+            tagged_records = @records.tagged_with(@params[:search].downcase)
+            ids = tagged_records
+              .map {|t| t[@resource.primary_key]}
+              .join(',')
+
+            conditions << "#{@resource.primary_key} IN (#{ids})" if ids.present?
+          end
+        end
+
         SchemaUtils.one_associations(@resource).map(&:name).each do |association|
           if @includes.include? association.to_sym
             resource = @resource.reflect_on_association(association.to_sym)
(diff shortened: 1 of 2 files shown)


SeyZ:
If you agree with the implementation, do you think it's easy to only allow "contains" filter on the UI @arnaudbesnier ?

arnaudbesnier:
The `tags` field seems to be save with a `String` type in Forest database (according to @lucasscariot) so I don't think it will be easy to do that.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -79,13 +79,29 @@ def search_param
...
+
         @params[:filter].each do |field, values|
-          next if association?(field)
+          # ActsAsTaggable
+          if acts_as_taggable?(field)
+            tagged_records = @records.tagged_with(values.tr('*', ''))
+
+            if @params[:filterType] == 'and'
+              @records = tagged_records
+            elsif @params[:filterType] == 'or'
```
arnaudbesnier: If you can do that for the filters, why not doing it this way (double request) for the search?

app/services/forest_liana/search_query_builder.rb
```diff
@@ -79,13 +79,29 @@ def search_param
...
+              ids = tagged_records
+                .map {|t| t[@resource.primary_key]}
+                .join(',')
+
+              conditions << "#{@resource.primary_key} IN (#{ids})"
```
arnaudbesnier: Shouldn't you escape the column name to prevent issues?

SeyZ:
[(2) Users want to search on a "tags" field created using the "acts_as_taggable" gem](https://trello.com/c/Vojc8RtZ/1421-2-users-want-to-search-on-a-tags-field-created-using-the-acts-as-taggable-gem)

app/services/forest_liana/search_query_builder.rb
```diff
@@ -34,6 +34,16 @@ def format_column_name(table_name, column_name)
       ForestLiana::AdapterHelper.format_column_name(table_name, column_name)
     end
 
+    def acts_as_taggable_query(tagged_records)
+      ids = tagged_records
+        .map {|t| t[@resource.primary_key]}
+        .join(',')
+
+      if ids.present?
```
arnaudbesnier: On a search, If there is no `ids` found (no record tagged with the search value), the liana crashes.

PR #152, opened 2017-05-29
[+] Filters - Add the not contains operator.

lucasscariot:


lucasscariot:
[(8) Users want to filter with the "not contain" operator](https://trello.com/c/1elR551d/221-8-users-want-to-filter-with-the-not-contain-operator)

PR #157, opened 2017-06-22
[*] Records Lists - Try to fix records display with specific hasOne associations

arnaudbesnier:


arnaudbesnier:
This is a bad fix / Breaks automatic tests

PR #266, opened 2018-06-22
[*] Server Start - Use namespace for ResourcesController to prevent conflicts

jonathantribouharet:
Hi,
I couldn't launch my server with forest-rails gem because the class `ControllerFactory` was using `ResourcesController` which is a common name, used in my case in another gem.
So I just add the namespace in your code and everything is working fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- Namespacing - Fix conflict with ResourcesController
 
 ## RELEASE 2.8.4 - 2018-06-21
 ### Changed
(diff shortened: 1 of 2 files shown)


arnaudbesnier:
We'll definitely merge this today!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 - Permissions - Fix automated permission for projects having multiple teams.
+- Namespacing - Fix conflict with ResourcesController
```
arnaudbesnier: You added the change in a released version (v2.8.4) but your fixed will be available in new gem version.
Can you create a new "Fixed" section just below "[Unreleased]" with your change description, like this:

 ```markdown
## [Unreleased]
### Fixed
- Namespacing - Fix conflict with ResourcesController
```

jonathantribouharet:
Done.

arnaudbesnier:
Thanks @jonathantribouharet for the contribution, it will be available in the next release!

🌲🌲🌲

Issue #252, opened 2018-04-03
[*] Search - Escaping issues / Possible SQL injection on searches?


blakewest:
## Expected behavior

I would expect that when passing in a string to search by, it would be properly escaped, and I would not get an error.

## Actual behavior

I see an error. Specifically, a syntax error when passing in a `'` to the search bar.

## Failure Logs
This image is from our Sentry logs of when this happened.
![image](https://user-images.githubusercontent.com/2524396/38272578-79ff0c5c-373e-11e8-9510-a37945a5752d.png)

Some other stuff from the stack trace:
![image](https://user-images.githubusercontent.com/2524396/38272738-f8a51254-373e-11e8-9be1-26576ffd07e9.png)

## Context
Our customer support admin was using forest and ran into this error. Is this possibly SQL injection?

* Package Version: 1.7.7
* Rails Version: 4.2
* Database Dialect: Postgres
* Database Version: 9.6

Uelb:
@blakewest I think you are right, I dug in the code to understand what was happening as I got the same issue, and it seems like it's caused by this line : 
https://github.com/ForestAdmin/forest-rails/blob/devel/app/services/forest_liana/search_query_builder.rb#L76
As you can see, there is no escaping whatsoever.

arnaudbesnier:
@blakewest , @Uelb this is fixed in the latest liana version (v2.11.4).

Thanks for the report!
🌲🌲🌲

PR #276, opened 2018-07-17
Use active record escaping pattern to avoid sql injection when search…

Uelb:
…ing for a record
This fixes https://github.com/ForestAdmin/forest-rails/issues/252 which is an important security issue.

## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Security - Avoid SQL Injection by escaping the search parameter when using active record escaping pattern.
 
 ## RELEASE 2.10.5 - 2018-07-11
 ### Fixed
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -12,17 +12,18 @@ module ForestLiana
     end
 
     def perform(resource)
+      @search = @params[:search]
       @resource = @records = resource
       @records = search_param
       @records = filter_param
       @records = has_many_filter
       @records = belongs_to_filter
 
-      if @params[:search]
+      if @search
         ForestLiana.schema_for_resource(@resource).fields.each do |field|
           if field.try(:[], :search)
             begin
-              @records = field[:search].call(@records, @params[:search])
+              @records = field[:search].call(@records, @search)
             rescue => exception
               FOREST_LOGGER.error "Cannot search properly on Smart Field :\n" \
                 "#{exception}"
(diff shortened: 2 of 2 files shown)


Uelb:
I haven't seen any implemented tests on what I changed, so I did not add them, I'm not sure what you expect on this part.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Security - Avoid SQL Injection by escaping the search parameter when using active record escaping pattern.
```
arnaudbesnier: Hey @Uelb , thanks for the contribution.

Can you provide an example of search value that could generate an SQL injection (for test purposes)?

Uelb: Actually, any value that has a quote : "'" would make the application fail, and if you add comments marks or stuff like that, it would make it possible to change the query. just try to search "L'opération" in a search bar for instance. Do you want me to put this information in the changelog ?

Uelb: (single quote)

arnaudbesnier: If I understand well, it is not a security issue then. I managed to reproduce the issue, you're right such search generate an internal error but I did not manage to do any SQL injection as it always generates a query having a bad syntax.

Uelb: I uninstalled forest from my app so I can't try anymore, but you can find more information here : https://fr.wikipedia.org/wiki/Injection_SQL

If you put something like '; ---', or something like that, it may produce a correct SQL query. It surely depends on the query generated, but the fact that characters are not escaped when sending a query is in itself a security issue that someone could exploit. Of course, it's not a major one here as the user has to be authenticated already but I do think it's security related.

Uelb: You could try (in a test database) :
"abc' ; DELETE FROM users ; ---" or something similar like
or examples here : https://www.w3schools.com/sql/sql_injection.asp
(Highly dangerous to try that btw)

arnaudbesnier: If you're using Postgresql, PG protects the database against such attack:
`PG::SyntaxError: ERROR:  cannot insert multiple commands into a prepared statement`

Didn't test it with SQLite, MySQL,... databases.

Uelb: Nice ! I did not know that, and what about just adding something like OR 1=1 to the query ? It would allow a user to access records he should not have access to no ?

arnaudbesnier:
Thanks for the contribution @Uelb !

🌲🌲🌲

PR #280, opened 2018-08-20
[*] Collections - Fixes potential behaviour issues of collections having a name that collides with query params keys

VincentMolinie:
## Pull Request checklist:

- [X] Write an explicit title for the Pull Request
- [X] Write changes made in the CHANGELOG.md
- [ ] Create automatic tests

### DIFF
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -50,7 +50,7 @@ module ForestLiana
     end
 
     def search_param
-      if (@search && !@search.empty?())
+      if @search
         conditions = []
 
         @resource.columns.each_with_index do |column, index|
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -50,7 +50,7 @@ module ForestLiana
     end
 
     def search_param
-      if @search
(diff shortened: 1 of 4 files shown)


app/services/forest_liana/search_query_builder.rb
```diff
@@ -50,7 +50,7 @@ def acts_as_taggable_query(tagged_records)
     end
 
     def search_param
-      if @search
+      if (@search && !@search.empty?())
```
arnaudbesnier: Can you explain the fix for the review in Trello or Github?
It is quite hard to follow you reflexion based on this line of code.

VincentMolinie: The issue is that when you want your collections, rails crash with an error `unknown function to_i` because search is actually empty (search = {}). So I just check if search is empty first before computing any search params

arnaudbesnier: Actually the issue is more generic than only the `search`. Any collection name that could interfere with query params can raise issues:

screen shot 2018-08-21 at 12 06 33

So `pages`, `searches`, `fields`, `sorts`, `timezones` are table name that could generate the same kind of error.
We need to find a way to prevent Rails (I feel that this is Rails that does that) to add `"collectionName" => {}` in the params.

PR #290, opened 2018-10-15
[*] Smart Views - Smart views can access nested relationship

VincentMolinie:
## Pull Request checklist:

- [X] Write an explicit title for the Pull Request
- [X] Write changes made in the CHANGELOG.md
- [ ] Create automatic tests
- [X] Test manually the implemented changes

### DIFF
--- a/app/controllers/forest_liana/application_controller.rb
+++ b/app/controllers/forest_liana/application_controller.rb
@@ -31,6 +31,31 @@ module ForestLiana
       @jwt_decoded_token
     end
 
+    def define_belongsTo_id(records, already_mapped_records = {})
+      records.each do |record|
+        if !already_mapped_records.key?(record.class.name)
+          already_mapped_records[record.class.name] = []
+        end
+        if !already_mapped_records[record.class.name].include? record.id
+          already_mapped_records[record.class.name] << record.id
+          @associations = record.class.reflect_on_all_associations
+          @associations.each do |association|
+            if [:belongs_to, :has_one].include?(association.macro) &&
+              !record[association.name] &&
+              !record.association(association.name).loaded?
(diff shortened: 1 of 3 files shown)


VincentMolinie:
[P2 - Plugify can't get his smart view to work properly](https://trello.com/c/zdhDFnIZ/2937-p2-plugify-cant-get-his-smart-view-to-work-properly)

VincentMolinie:
To add more context about how JSON API works:
First thing we need to know is that if in a relationships, there is a link specified even if the data are retrieved in the included, Ember Data will do a GET request on the url. So link must be here ONLY if the data is not included. Then where there is data in relationships `{ data: { id: '1', type: 'student'}}` and a link the data will always be the first one to be called. If none of them is specified, the relation is considered as null.

app/controllers/forest_liana/application_controller.rb
```diff
@@ -31,6 +31,31 @@ def forest_user
       @jwt_decoded_token
     end
 
+    def define_belongsTo_id(records, already_mapped_records = {})
```
VincentMolinie: This function allow to add a `{ data: { id: 1 } }` on all belongsTo which are not retrieved from the previous query. For that we just instantiate an empty object with an id so that the serializer put the data

app/controllers/forest_liana/application_controller.rb
```diff
@@ -31,6 +31,31 @@ def forest_user
...
+      records.each do |record|
+        if !already_mapped_records.key?(record.class.name)
+          already_mapped_records[record.class.name] = []
+        end
+        if !already_mapped_records[record.class.name].include? record.id
```
VincentMolinie: we make sure the record has not already been "defined" to avoid infinite loop

arnaudbesnier:
Let's keep this for later. We need stability now and I don't want to introduce unecessary feature. Especially on the liana side.

Issue #178, opened 2017-08-11
Support for Error Logging instrumentation


karensg:
Hi guys,

Whenever there is an issue/error in the Forest code, the error is rescued and a 404 is returned. This is an unpleasant behavior because the actual errors are not rescued by the error logging software. We use [AppSignal](https://appsignal.com/)

All production applications are using some sort of  error logging tools, so it would be great to not swallow errors but properly log them in order to monitor the Forest performance.

Let me know what are your thoughts about it.

Karens

arnaudbesnier:
Hi @karensg,

Thanks for the feedback!
I am not sure to understand. As far as I know, we don't rescue any error happening in Forest with 404, only specific errors (routing errors and resources not found).

Can you provide an example of code that generates such behaviour?

Thank you

karensg:
So the problem I encountered was that because of the rescues which are used in your code base, the logging tools don't catch the errors as it looks like the error is handled already. It is very inconvenient as you are not aware of the things which go wrong.
https://github.com/ForestAdmin/forest-rails/blob/20f639579befb1fea1065fc64b51daa91f2bb986/app/controllers/forest_liana/resources_controller.rb

PR #325, opened 2019-08-01
Support third party error logging via simple instrumentation

plcstevens:
Hi,

Being able to report exceptions to a third party tool is a common fallback in a lot of Ruby and Ruby on Rails projects. In our case we use Sentry to manage error fallout and monitoring, but we discovered recently that a large amount of Forest related errors were being swallowed and reported only in the logs. This issue has been raised previously here: https://github.com/ForestAdmin/forest-rails/issues/178 but without any follow up.

This MR attempts to provide a very simple interface for bubbling errors out of the Forest application. It is set explicitly whenever Forest rescue's an error but does not raise a new one. By default it will do nothing with the exception, and it is up to the user to override that default behaviour.

Here are the ways you'd report the error to either Sentry or Airbrake, two very popular error monitoring tools.

```ruby
# Sentry
ForestLiana.error_handler = ->(exception) { Raven.capture_exception(exception) }
# Airbrake
ForestLiana.error_handler = ->(exception) { Airbrake.notify(exception) }
```

## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- Functionality - Allow users to define their own error handling instrumentation for exceptions raised/rescued throughout the application.
+
 ## RELEASE 3.2.0 - 2019-07-22
 ### Added
 - Column Types - Support HSTORE column type.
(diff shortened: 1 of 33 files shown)


ramongtx:
+1

ramongtx:
+1

PR #327, opened 2019-09-09
Add initialisation of configuration directories

purinkle:
We had hardcoded the directory used for configuring more data models. Hardcoding this value made it difficult to set a custom directory to define these files. It also caused `zeitwerk` to throw a warning when trying to autoload the extra classes. We added a `config_dirs` accessor to the configuration module. Adding this attribute allows us to set this directory ourselves.

## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Configuration - Add initialisation of configuration directories
+
 ### Changed
 - Readme - Add a community section.
 - Readme - Remove the Licence section as it is already accessible in the Github page header.
--- a/lib/forest_liana.rb
+++ b/lib/forest_liana.rb
@@ -37,8 +37,18 @@ module ForestLiana
   self.names_overriden = {}
   self.meta = {}
 
+  @config_dir = 'lib/forest_liana/**/*.rb'
+
   # TODO: Remove once lianas prior to 2.0.0 are not supported anymore.
   self.names_old_overriden = {}
+
+  def self.config_dir=(config_dir)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Changed
 - Readme - Add a community section.
 - Readme - Remove the Licence section as it is already accessible in the Github page header.
+- Configuration - Add initialisation of configuration directories
```
arnaudbesnier: I think we can consider this as a feature, so it should be in the "Added" section:
```markdown
### Added
- Configurations - Users can specify the directory for Forest Smart Implementation.

### Changed
...
```

lib/forest_liana.rb
```diff
@@ -37,9 +37,23 @@ module UserSpace
...
+    @config_dirs = (dirs || []).map { |dir| Rails.root.join(dir) }
+  end
+
+  def self.config_dirs
+    [
+      Rails.root.join('lib/forest/**/*.rb'),
+      Rails.root.join('lib/forest_liana/**/*.rb'),
+      *@config_dirs
+    ]
+  end
```
arnaudbesnier: Everything looks great.

For consistency with other lianas/stacks, could we consider having a single repository instead of an array of repositories?
I also think it would be a better idea as it forces developers to chose a single repository to write their Forest Admin customization.
What do you think?

CHANGELOG.md
purinkle: Resolved in 5005b94

lib/forest_liana.rb
purinkle: Resolved in 5005b94

PR #332, opened 2019-11-04
Fix: Check serializer in ForestLiana::UserSpace module only

arambert:
The app could already have a PostSerializer (for instance) globally defined (for instance if the app is using JSON serializers outside of ForestAdmin).
When this is the case, this gem tries to remove the constant (`ForestLiana::UserSpace.send(:remove_const, serializer_name)`) from ForestLiana::UserSpace module but 1/ it isn't there and 2/ it should not remove it.

https://ruby-doc.org/core-2.6.4/Module.html#method-i-const_defined-3F

## Pull Request checklist:

- [ ] Write an explicit title for the Pull Request
- [ ] Write changes made in the CHANGELOG.md
- [ ] Create automatic tests

### DIFF
--- a/app/serializers/forest_liana/serializer_factory.rb
+++ b/app/serializers/forest_liana/serializer_factory.rb
@@ -6,7 +6,7 @@ module ForestLiana
     def self.define_serializer(active_record_class, serializer)
       serializer_name = self.build_serializer_name(active_record_class)
 
-      if ForestLiana::UserSpace.const_defined?(serializer_name)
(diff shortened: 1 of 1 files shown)


rchasman:
I am getting the same NameError in my project, making it impossible to upgrade to 4.1. After digging around got to this fix myself as well! Thanks for making a PR, hope this is merged this soon!

arnaudbesnier:
Thank you @arambert , @rchasman your patch is available in the brand new `v4.1.2`.

🌲🌲🌲

arambert:
Thank you!!! 🎉

PR #342, opened 2020-01-20
[*] Technical - Fix CI build

DrRaider:
Currently, Travis pulls the latest version of the bundler, whereas in our gemfile.lock, it is specified that the `1.17.3` is required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: ruby
 before_install:
-  -  gem install bundler -v '1.17.3'
+  - gem install bundler -v '1.17.3'
 rvm:
   - 2.3.4
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,8 @@
 sudo: false
 language: ruby
 before_install:
-  - gem update bundler
+  -  gem uninstall bundler
+  -  gem install bundler -v '1.17.3'
```
arnaudbesnier: The changelog is missing to describe your intension ;)

.travis.yml
```diff
@@ -1,7 +1,8 @@
 sudo: false
 language: ruby
 before_install:
-  - gem update bundler
+  -  gem uninstall bundler
```
arnaudbesnier: Is the `uninstall` command mandatory?

DrRaider: Seems to work without it

.travis.yml
DrRaider: Fixed

PR #345, opened 2020-02-24
[+] Resources getter - Add a get_ids_from_request method to get all models IDs given a query or an ID list

rap2hpoutre:
Add "select all records" mode for rails liana. See: https://app.clickup.com/t/3rhj80

## Usage in smart actions

```ruby
class Forest::ActorsController < ForestLiana::ApplicationController
  def mark_as_live
    actor_ids = ForestLiana::ResourcesGetter.get_ids_from_request(params)

    # do something.

    render json: { success: 'Data successfuly imported!' }
  end
end
```

## How to test

You have ton use this branch as a dependency in `Gemfile`
```
gem 'forest_liana', path: '../forest-rails'
```

Then you have to update `forestadmin` [liana feature file](https://github.com/ForestAdmin/forestadmin/blob/5c97aa9392c278b9aed616f1ea60086c7e11f4bf/app/services/liana-features.js) to enable "select all" feature:

```javascript
{
  name: 'tableViewSelectAllRecords',
  minimalVersion: {
    'forest-express-sequelize': '6.0.0',
    'forest-express-mongoose': '6.0.0',
    'forest-rails': '4.2.0', // <- use this version
  },
}
```

## TODO list
 - [x] implement `get_ids_from_request`
   - [x] It should still work for id lists
   - [x] it should work for all records
   - [x] it should consider excluded ids
   - [x] it should work for nested collections
- [x] implement delete bulk
   - [x] it should work for nested collections (disociation
   - [x] it should work for nested collections (deletion)
   - [x] it should work for collection

## Pull Request checklist:

- [ ] Create automatic tests

## Important note

⚠️  The code I wrote is ugly, inconsistant, tinkered and flawes: feel free to suggest (or even commit) better code.

### DIFF
--- a/app/controllers/forest_liana/resources_controller.rb
+++ b/app/controllers/forest_liana/resources_controller.rb
@@ -137,23 +137,28 @@ module ForestLiana
     end
 
     def destroy
-      begin
-        checker = ForestLiana::PermissionsChecker.new(@resource, 'delete', @rendering_id)
-        return head :forbidden unless checker.is_authorized?
+      checker = ForestLiana::PermissionsChecker.new(@resource, 'delete', @rendering_id)
+      return head :forbidden unless checker.is_authorized?
 
-        # NOTICE: destroy one or many records.
-        if (params[:id] && params.dig('data', 'attributes') == nil)
-          @resource.destroy(params[:id]) if @resource.exists?(params[:id])
-        else 
-          ids = ForestLiana::ResourcesGetter.get_ids_from_request(params)
-          @resource.destroy(ids)
-        end
-        
-        head :no_content
-      rescue => error
-        FOREST_LOGGER.error "Record Destroy error: #{error}\n#{format_stacktrace(error)}"
-        internal_server_error
-      end
+      @resource.destroy(params[:id]) if @resource.exists?(params[:id])
+
+      head :no_content
+    rescue => error
+      FOREST_LOGGER.error "Record Destroy error: #{error}\n#{format_stacktrace(error)}"
+      internal_server_error
+    end
+
+    def destroy_many
+      checker = ForestLiana::PermissionsChecker.new(@resource, 'delete', @rendering_id)
+      return head :forbidden unless checker.is_authorized?
+
+      ids = ForestLiana::ResourcesGetter.get_ids_from_request(params)
+      @resource.destroy(ids) if ids&.any?
+
+      head :no_content
+    rescue => error
+      FOREST_LOGGER.error "Records Destroy error: #{error}\n#{format_stacktrace(error)}"
+      internal_server_error
     end
 
     private
--- a/app/controllers/forest_liana/resources_controller.rb
+++ b/app/controllers/forest_liana/resources_controller.rb
@@ -141,7 +141,14 @@ module ForestLiana
         checker = ForestLiana::PermissionsChecker.new(@resource, 'delete', @rendering_id)
         return head :forbidden unless checker.is_authorized?
 
-        @resource.destroy(params[:id]) if @resource.exists?(params[:id])
+        # NOTICE: destroy one or many records.
+        if (params[:id] && params.dig('data', 'attributes') == nil)
+          @resource.destroy(params[:id]) if @resource.exists?(params[:id])
+        else 
+          ids = ForestLiana::ResourcesGetter.get_ids_from_request(params)
+          @resource.destroy(ids)
+        end
+        
         head :no_content
       rescue => error
         FOREST_LOGGER.error "Record Destroy error: #{error}\n#{format_stacktrace(error)}"
--- a/app/services/forest_liana/has_many_dissociator.rb
+++ b/app/services/forest_liana/has_many_dissociator.rb
@@ -14,10 +14,12 @@ module ForestLiana
 
       remove_association = !@with_deletion || @association.macro == :has_and_belongs_to_many
 
-      if @data.is_a?(Array) && @data.dig('attributes').nil?
+      if @data.is_a?(Array)
         record_ids = @data.map { |record| record[:id] }
-      elsif !@data.dig('attributes').nil?
+      elsif @data.dig('attributes').present?
         record_ids = ForestLiana::ResourcesGetter.get_ids_from_request(@params)
+      else
+        record_ids = Array.new
       end
 
       if !record_ids.nil? && record_ids.any?
--- a/app/services/forest_liana/has_many_dissociator.rb
+++ b/app/services/forest_liana/has_many_dissociator.rb
@@ -14,20 +14,31 @@ module ForestLiana
 
       remove_association = !@with_deletion || @association.macro == :has_and_belongs_to_many
 
+      record_ids = Array.new
+
       if remove_association
-        if @data.is_a?(Array)
-          @data.each do |record_deleted|
-            associated_records.delete(@association.klass.find(record_deleted[:id]))
+        if @data.is_a?(Array) && @data.dig('attributes') == nil
+          record_ids = @data.map { |record| record[:id] }
+        elsif @data.dig('attributes') != nil
+          record_ids = ForestLiana::ResourcesGetter.get_ids_from_request(@params)
+        end
+        if record_ids.any?
+          record_ids.each do |id|
+            associated_records.delete(@association.klass.find(id))
           end
         end
       end
 
       if @with_deletion
-        if @data.is_a?(Array)
+        if @data.is_a?(Array) && @data.dig('attributes') == nil
           record_ids = @data.map { |record| record[:id] }
+        elsif @data.dig('attributes') != nil
+          record_ids = ForestLiana::ResourcesGetter.get_ids_from_request(@params)
+        end
+        if record_ids.any?
           record_ids = record_ids.select { |record_id| @association.klass.exists?(record_id) }
           @association.klass.destroy(record_ids)
-        end
+        end 
       end
     end
   end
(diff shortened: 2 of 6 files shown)


app/controllers/forest_liana/resources_controller.rb
```diff
@@ -141,7 +141,14 @@ def destroy
...
         return head :forbidden unless checker.is_authorized?
 
-        @resource.destroy(params[:id]) if @resource.exists?(params[:id])
+        # NOTICE: destroy one or many records.
+        if (params[:id] && params.dig('data', 'attributes') == nil)
+          @resource.destroy(params[:id]) if @resource.exists?(params[:id])
+        else 
+          ids = ForestLiana::ResourcesGetter.get_ids_from_request(params)
+          @resource.destroy(ids)
+        end
```
arnaudbesnier: [Thinking] I wonder if it is a good idea to have 1 single controller for 2 routes.
I know that you'll factorize some logic but:
- it makes the code harder to read
- the `rescue` error log is the same for single and multiple destroy so the context will be hard to get in case of error.

Maybe `class ForestLiana::Router` should be adapted to create a dedicated controller for multiple destroy.

app/services/forest_liana/has_many_dissociator.rb
```diff
@@ -14,20 +14,31 @@ def perform
...
         end
       end
 
       if @with_deletion
-        if @data.is_a?(Array)
+        if @data.is_a?(Array) && @data.dig('attributes') == nil
           record_ids = @data.map { |record| record[:id] }
+        elsif @data.dig('attributes') != nil
+          record_ids = ForestLiana::ResourcesGetter.get_ids_from_request(@params)
+        end
```
arnaudbesnier: [Maintenance] Aren't the lines 33-37 a duplicate of lines 20-24?

app/services/forest_liana/has_many_dissociator.rb
```diff
@@ -14,20 +14,31 @@ def perform
 
       remove_association = !@with_deletion || @association.macro == :has_and_belongs_to_many
 
+      record_ids = Array.new
```
arnaudbesnier: [Question] Not really a convention in our Ruby code to instantiate variables that will be affected in the future.
Maybe you won't have issue to remove this if you use `if !record_ids.nil? && record_ids.any?` in the code below.

app/services/forest_liana/resources_getter.rb
```diff
@@ -19,6 +19,49 @@ def initialize(resource, params)
       prepare_query()
     end
 
+    def self.get_ids_from_request(params)
+      attributes = params.dig('data', 'attributes')
+      has_body_attributes = attributes != nil
+      is_select_all_records_query = has_body_attributes && attributes[:all_records] === true
```
arnaudbesnier: [Syntax] is the triple equal really necessary (`===` => `==`) ?

app/services/forest_liana/resources_getter.rb
```diff
@@ -19,6 +19,49 @@ def initialize(resource, params)
...
+    
+    
+      # NOTICE: If it is not a "select all records" query and it receives a list of ID, return list of ID.
+      return attributes[:ids] if (!is_select_all_records_query && attributes[:ids])
+
+      # NOTICE: If it is a "select all records" we have to perform query to build ID list.
+      ids = Array.new
+
+      # NOTICE: Initialize actual resources getter (could either a HasManyGetter or a ResourcesGetter).
+      isRelatedData = attributes[:parent_collection_id] &&
```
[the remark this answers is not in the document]
arnaudbesnier: [Syntax] If allowed, would it be better like that?
```
is_related_data = attributes[:parent_collection_id]
  && attributes[:parent_collection_name]
  && attributes[:parent_association_name]
```

app/services/forest_liana/resources_getter.rb
```diff
@@ -19,6 +19,49 @@ def initialize(resource, params)
...
+          association_name: attributes[:parent_association_name], 
+          collection: parent_collection_name
+        }))
+      else
+        collection_name = attributes[:collection_name]
+        model = ForestLiana::SchemaUtils.find_model_from_collection_name(collection_name)
+        resources_getter = ForestLiana::ResourcesGetter.new(model, attributes)
+      end
+
+      # NOTICE: build ID list.
```
arnaudbesnier: A "s" after the "ID" mention would help to understand the multiple context.
```suggestion
      # NOTICE: build IDs list.
```

app/services/forest_liana/resources_getter.rb
```diff
@@ -19,6 +19,49 @@ def initialize(resource, params)
...
+        attributes[:parent_collection_name] &&
+        attributes[:parent_association_name]
+      if (isRelatedData)
+        parent_collection_name = attributes[:parent_collection_name]
+        parent_model = ForestLiana::SchemaUtils.find_model_from_collection_name(parent_collection_name)
+        model = parent_model.reflect_on_association(attributes[:parent_association_name].try(:to_sym))
+        resources_getter = ForestLiana::HasManyGetter.new(parent_model, model, attributes.merge({ 
+          id: attributes[:parent_collection_id], 
+          association_name: attributes[:parent_association_name], 
+          collection: parent_collection_name
```
arnaudbesnier:
```ruby
collection: parent_collection_name,
id: attributes[:parent_collection_id], 
association_name: attributes[:parent_association_name], 
```
This order would improve the readability.

app/services/forest_liana/resources_getter.rb
```diff
@@ -19,6 +19,49 @@ def initialize(resource, params)
...
+        resources_getter = ForestLiana::ResourcesGetter.new(model, attributes)
+      end
+
+      # NOTICE: build ID list.
+      resources_getter.query_for_batch.find_in_batches() do |records|
+        ids += records.map { |record| record.id }
+      end
+
+      # NOTICE: remove excluded IDs.
+      ids_excluded = (attributes[:all_records_ids_excluded]).map { |id_excluded| id_excluded.to_s }
```
arnaudbesnier: [Question] Is this line really necessary?
I am just wondering if we can simplify this code.

app/services/forest_liana/resources_getter.rb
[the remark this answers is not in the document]
rap2hpoutre: > If allowed, would it be better like that?

Unfortunately, it is not allowed :/

app/services/forest_liana/resources_getter.rb
rap2hpoutre: Good catch! 

I used the same logic in forest-express: force `ids_exluded` and `ids` to strings before comparing them (to avoid unexpected behavior with mongo's `ObjectId`s, `int`, `uuid`, etc.). **It's not necessary**(?), but testing and verifying various cases was not that simple and I'm not confident about being aware of all cases. It's overprotective, but I prefer use this to be consistant with `forest-express`, not trusting data that comes from frontend and avoid unexpected deletion. Furthermore, I don't think it's too expensive. So I guess it's OK, still your question was legit!

SteveBunlon:
First of all, I had to edit the `has_many_dissaciator.rb` file cause a runtime error prevented me to test : 
Screenshot 2020-02-28 at 17 11 05

Here is the code I used to fix : 
```
    def perform
      @record = @resource.find(@params[:id])
      associated_records = @resource.find(@params[:id]).send(@association.name)

      remove_association = !@with_deletion || @association.macro == :has_and_belongs_to_many

      if @data.is_a?(Array)
        record_ids = @data.map { |record| record[:id] }
      elsif @data.is_a?(Hash) && !@data.dig('attributes').nil?
        record_ids = ForestLiana::ResourcesGetter.get_ids_from_request(@params)
      else
        record_ids = Array.new
      end

      if !record_ids.nil? && record_ids.any?
        if remove_association
          record_ids.each do |id|
            associated_records.delete(@association.klass.find(id))
          end
        end

        if @with_deletion
          record_ids = record_ids.select { |record_id| @association.klass.exists?(record_id) }
          @association.klass.destroy(record_ids)
        end
      end
    end
```

SteveBunlon:
I have performed a huge test batterie for this one.

⚠️ Please not that I used the code below to perform tests on dissociations. 

Here are the results : 

Screenshot 2020-02-28 at 18 08 26

SteveBunlon:
Also I noticed a bug why selected all records front. You told me that this is not related to this PR, but I prefer to log it there as we still got unexpected errors (so it might be related). 

When I filter the related data, the count is not correct (see at the bottom right of the screen) : 
Screenshot 2020-02-28 at 18 14 49

Also, if I try to select all, the count is not correct either (check the toast): 
Screenshot 2020-02-28 at 18 14 59

rap2hpoutre:
Thank you so much for your review! 

> ⚠️ Please not that I used the code below to perform tests on dissociations.

I just spotted the problem in the code we crafted to make it work. I had to remove the `@data.is_a?(Hash)` since data is not technically a hash (it's a kind of inherited hash), so I had to replace this:

```ruby
elsif @data.is_a?(Hash) && !@data.dig('attributes').nil?
```

by that:

```ruby
elsif !@data.dig('attributes').nil?
```

See the commit: https://github.com/ForestAdmin/forest-rails/pull/345/commits/9fa16d66c4c23e527f25e715a9051d9e7871b73d

> Also I noticed a bug why selected all records front. You told me that this is not related to this PR, but I prefer to log it there as we still got unexpected errors (so it might be related).

I suggest to consider it as another issue. I will try to tackle it for the next hour: if I found and fix the issue, I will submit it in a new PR, otherwise, I will open a ticket.

PR #348, opened 2020-03-02
[*] Has Many Relationships - Fix records count, i.e consider filters when counting

rap2hpoutre:
## Description

See @SteveBunlon's comment: https://github.com/ForestAdmin/forest-rails/pull/345#issuecomment-592613766

> When I filter the related data, the count is not correct (see at the bottom right of the screen)

This PR consider filters before counting related data.

## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -19,11 +19,11 @@ module ForestLiana
     end
 
     def perform
-      @records = @search_query_builder.perform(@records)
+      @records
     end
 
     def count
-      @records_count = @search_query_builder.perform(@records).count
+      @records_count = @records.count
     end
 
     def query_for_batch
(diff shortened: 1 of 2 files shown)


app/services/forest_liana/has_many_getter.rb
```diff
@@ -23,7 +23,7 @@ def perform
     end
 
     def count
-      @records_count = @records.count
+      @records_count = @search_query_builder.perform(@records).count
```
arnaudbesnier: Looks cool, shouldn't we do this in `query_for_batch` and `records` too?
If so, maybe the `@search_query_builder.perform` should be done `prepare_query`?

rap2hpoutre: Nice catch! Fixed!

SteveBunlon:
Tested this manually, works like a charm 👍 ⭐️

PR #361, opened 2020-05-28
feat(scope): validate scope context on list/count request

jeffladiray:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/app/services/forest_liana/scope_validator.rb
+++ b/app/services/forest_liana/scope_validator.rb
@@ -1,0 +1,72 @@
+module ForestLiana
+  class ScopeValidator
+    def initialize(scope_permissions, scope_values)
+      @scope_filters = scope_permissions
+      @dynamic_scope_variables = scope_values
+    end
+
+    def is_scope_in_request?(scope_request)
+      begin
+        filters = JSON.parse(scope_request[:filters])
+      rescue JSON::ParserError
+        raise ForestLiana::Errors::HTTP422Error.new('Invalid filters JSON format')
+      end
+      @computed_scope = compute_condition_filters_from_scope(scope_request[:user_id])
+      tagged_scope_filters = validate(filters)
+      return tagged_scope_filters != nil if @scope_filters['conditions'].length == 1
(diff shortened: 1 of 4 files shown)


app/services/forest_liana/scope_validator.rb
```diff
@@ -0,0 +1,94 @@
...
+      # NOTICE: Remove conditions that are not from the scope
+      filtered_conditions = node['conditions'].map { |condition| 
+        search_scope_aggregation(condition)
+      }.select { |condition|
+        condition
+      }
+
+      # NOTICE: If there is only one condition filter left and it's current aggregator is
+      #         an "and", this condition filter is the searched scope
+      if (filtered_conditions.length === 1 &&
```
larcin, line 61: [MISC] As stated [here](https://stackoverflow.com/questions/3422223/vs-in-ruby), you should avoid the `===` comparison operator as it does not have the same meaning as in JS

```suggestion
      if (filtered_conditions.length == 1 && 
```

app/services/forest_liana/scope_validator.rb
```diff
@@ -0,0 +1,94 @@
...
+      if (filtered_conditions.length === 1 && 
+        filtered_conditions.first.is_a?(Hash) &&
+        filtered_conditions.first.include?(:aggregator) &&
+        node['aggregator'] == 'and')
+        return filtered_conditions.first
+      end
+
+      # NOTICE: Otherwise, validate if the current node is the scope and return nil
+      #         if it's not
+      return (filtered_conditions.length === @scope_filters['conditions'].length &&
```
larcin, line 70:
```suggestion
      return (filtered_conditions.length == @scope_filters['conditions'].length && 
```

test/services/forest_liana/scope_validator_test.rb
```diff
@@ -0,0 +1,174 @@
...
+        ).is_scope_in_request?({
+          user_id: '1',
+          filters: JSON.generate({
+            aggregator: 'and',
+            conditions: [
+              { field: 'name', value: 'john', operator: 'equal' },
+              { field: 'price', value: '2500', operator: 'equal' }              
+            ]
+          })
+        })
```
larcin, line 20: [MISC] Out of readability concerns, could you split the ScopeValidator instance creation and the call to `is_scope_in_request?` ?

PR #371, opened 2020-08-03
fix(vulnerability): patch a potential vulnerability updating lodash dependency to version 4.17.19

amessinger:
Since lodash is a dependency of dependencies in this project, I've used the ["resolutions"](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) field to enforce the use of the fixing versions (the bump is a patch so nothing breaking here).

## Pull Request checklist:

- [X] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Create automatic tests
- [X] Wonder if you can improve the existing code

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,8 @@
     "husky": "4.2.5",
     "semantic-release": "17.0.7",
     "semantic-release-slack-bot": "1.6.2"
+  },
+  "resolutions": {
+    "**/lodash": "4.17.19"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,32 +3,25 @@
 
 
 "@babel/code-frame@^7.0.0":
-  version "7.8.3"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -28,5 +28,8 @@
     "husky": "4.2.5",
     "semantic-release": "17.0.7",
     "semantic-release-slack-bot": "1.6.2"
+  },
+  "resolutions": {
+    "**/lodash": "4.17.19"
```
arnaudbesnier, line 33: Really not a fan of this, I don't understand how we can confidently bypass dependencies rules.
This feature should be used only in case of "emergency" in my opinion.

arnaudbesnier: Here is my update proposal to get rid of the lodash vulnerability:
https://github.com/ForestAdmin/forest-rails/pull/372

amessinger: I went for the "quickest" solution but your approach is by far the cleanest. Let's go for it.

amessinger:
Superseeded by https://github.com/ForestAdmin/forest-rails/pull/372

PR #382, opened 2020-11-25
feat: handle hooks

GuillaumeCisco:
## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/lib/forest_liana/bootstrapper.rb
+++ b/lib/forest_liana/bootstrapper.rb
@@ -56,10 +56,10 @@ module ForestLiana
 
         @collections_sent.each do |collection|
           collection['actions'].each do |action|
-            collection = get_collection(collection['name'])
-            a = get_action(collection, action['name'])
-            load = !a.hooks.nil? ? a.hooks.key?(:load) && a.hooks[:load].is_a?(Proc) : false
-            change = !a.hooks.nil? ? (a.hooks.key?(:change) && a.hooks[:change].is_a?(Hash) ? a.hooks[:change].keys : []) : []
+            c = get_collection(collection['name'])
+            a = get_action(c, action['name'])
+            load = !a.hooks.nil? && a.hooks.key?(:load) && a.hooks[:load].is_a?(Proc)
+            change = !a.hooks.nil? && a.hooks.key?(:change) && a.hooks[:change].is_a?(Hash) ? a.hooks[:change].keys : []
             action['hooks'] = {:load => load, :change => change}
           end
         end
--- a/lib/forest_liana/bootstrapper.rb
+++ b/lib/forest_liana/bootstrapper.rb
@@ -38,6 +38,14 @@ module ForestLiana
 
     private
 
+    def get_collection(collection_name)
+      ForestLiana.apimap.find { |collection| collection.name.to_s == collection_name }
+    end
+
+    def get_action(collection, action_name)
+      collection.actions.find {|action| action.name == action_name}
+    end
+
     def generate_apimap
       create_apimap
       require_lib_forest_liana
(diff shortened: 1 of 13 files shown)


lib/forest_liana/bootstrapper.rb
```diff
@@ -38,13 +38,34 @@ def display_apimap
...
+            unless action['hooks'].nil?
+              collection = get_collection(collection['name'])
+              a = get_action(collection, action['name'])
+              load = a.hooks.key?(:load) && a.hooks[:load].is_a?(Proc)
+              change = a.hooks.key?(:change) && a.hooks[:change].is_a?(Hash) ? a.hooks[:change].keys : []
+              action['hooks'] = {:load => load, :change => change}
+            end
+          end
+        end
+
```
GuillaumeCisco, line 68: I'm not sure if it is the right place for doing this business code, but I did not find another way :/

rap2hpoutre:
### How to use on `forest-live-demo-rails`

Add this in `lib/forest_liana/collections/customer.rb`:

```ruby
class Forest::Customer
  include ForestLiana::Collection

  collection :Customer

  action 'Charge credit card', type: 'single', fields: [{
    field: 'amount',
    is_required: true,
    description: 'The amount (USD) to charge the credit card. Example: 42.50',
    type: 'Number'
  }, {
    field: 'description',
    is_required: true,
    description: 'Explain the reason why you want to charge manually the customer here',
    type: 'String'
  },{
    field: 'stripe_id',
    type: 'String',
  }], hooks: {
    :load => -> (context){
      context[:fields]['amount'][:value] = 1000
      context[:fields]['description'][:value] = 'foo'
      return context[:fields]
    },
    :change => {
        'amount' => -> (context){
           context[:fields]['amount'][:value] = 2000
           return context[:fields]
        },
        'description' => -> (context){
          context[:fields]['description'][:value] = 'bar'
          return context[:fields]
        }
    }
}

  action 'Generate invoice', download: true

  set_fullname = lambda do |user_params, fullname|
    fullname = fullname.split
    user_params[:firstname] = fullname.first
    user_params[:lastname] = fullname.last

    # Returns a hash of the updated values you want to persist.
    user_params
  end

  search_fullname = lambda do |query, search|
    firstname, lastname = search.split

    # Injects your new filter into the WHERE clause.
    query.where_clause.send(:predicates)[0] << " OR (firstname = '#{firstname}' AND lastname = '#{lastname}')"

    query
  end

  field :fullname, type: 'String', set: set_fullname, search: search_fullname do
    "#{object.firstname} #{object.lastname}"
  end

  field :full_address, type: 'String' do
    address = Address.find_by(customer_id: object.id)
    "#{address[:address_line_1]} #{address[:address_line_2]} #{address[:address_city]} #{address[:country]}"
  end

  field :age, type: 'Number' do
    ((Time.zone.now - object.birth_date.to_time) / 1.year.seconds).floor if object.birth_date
  end
end

```

rap2hpoutre:
TODO: handle record (code can be improved)

```ruby
    def get_record
      model = ForestLiana::SchemaUtils.find_model_from_collection_name(params[:collectionName])
      getter = ForestLiana::ResourceGetter.new(model, {:id => params[:recordIds][0]})
      getter.perform
      getter.record
    end
```

and

```ruby
 {:record => get_record, :fields => fields}
```

rap2hpoutre (commented):
Super 🚀 

There is one difference with forest-express in forestadmin schema:

```
// forest-express:
      "hooks": {
        "load": false,
        "change": []
      }

// forest-rails:
      "hooks": null,
```

According to the EPIC, there should be a default value ([source](https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/eJGoNND3JQ#8397b239)) 

Not sure it's really a problem though. WDYT?

GuillaumeCisco:
> Super 🚀
> 
> There is one difference with forest-express in forestadmin schema:
> 
> ```
> // forest-express:
>       "hooks": {
>         "load": false,
>         "change": []
>       }
> 
> // forest-rails:
>       "hooks": null,
> ```
> 
> According to the EPIC, there should be a default value ([source](https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/eJGoNND3JQ#8397b239))
> 
> Not sure it's really a problem though. WDYT?

Can you tell me if it is ok now?

rap2hpoutre (approved):
🎉 It works!

PR #414, opened 2021-01-20
fix(smart-action-hook): value injected to an enum field of type is now correctly handled

SteveBunlon:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/spec/requests/actions_controller_spec.rb
+++ b/spec/requests/actions_controller_spec.rb
@@ -89,7 +89,7 @@ describe 'Requesting Actions routes', :type => :request  do
     }
     enums_action_definition = {
       name: 'enums_action',
-      fields: [foo, enum, multiple_enum],
+      fields: [foo, enum],
       hooks: {
         :change => {
           'foo' => -> (context) {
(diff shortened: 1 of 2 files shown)


spec/requests/actions_controller_spec.rb
```diff
@@ -84,7 +89,7 @@
     }
     enums_action_definition = {
       name: 'enums_action',
-      fields: [foo, enum],
+      fields: [foo, enum, multiple_enum],
```
GuillaumeCisco, line 92: Why is `multiple_enum` in `enums_action_definition`?
Can we keep it separated?

`enums_action_definition` with `fields: [foo, enum],`
and
`multiple_enums_action_definition` with `fields: [foo, multiple_enum],`

Would be easier to read and review :)

GuillaumeCisco (commented):
Great work! 🦖 
I absolutely love the `-` between arrays for intersection in ruby ❤️ 

I added a comment for dealing with tests ;)

spec/requests/actions_controller_spec.rb
SteveBunlon: True ! I'm changing this

GuillaumeCisco (approved):
That's perfect now! ✨

PR #426, opened 2021-02-19
fix: display a warning on missing association

jeffladiray:
## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/app/services/forest_liana/schema_utils.rb
+++ b/app/services/forest_liana/schema_utils.rb
@@ -2,17 +2,14 @@ module ForestLiana
   class SchemaUtils
 
     def self.associations(active_record_class)
-      associations = active_record_class.reflect_on_all_associations.select do |association|
-        next if association.options[:polymorphic] && !is_active_type?(association.klass)
+      active_record_class.reflect_on_all_associations.select do |association|
         begin
-          association.klass
+          !polymorphic?(association) && !is_active_type?(association.klass)
         rescue
           FOREST_LOGGER.warn "Unknown association #{association.name} on class #{active_record_class.name}"
+          false
         end
-        nil
       end
-
-      associations.compact
     end
 
     def self.one_associations(active_record_class)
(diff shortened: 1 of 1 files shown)


app/services/forest_liana/schema_utils.rb
```diff
@@ -2,9 +2,14 @@ module ForestLiana
   class SchemaUtils
 
     def self.associations(active_record_class)
-      active_record_class
-        .reflect_on_all_associations
-        .select { |association| !polymorphic?(association) && !is_active_type?(association.klass) }
+      associations = active_record_class.reflect_on_all_associations.select do |association|
+        next if association.options[:polymorphic] && !is_active_type?(association.klass)
```
arnaudbesnier, line 6: I am not sure you should apply directly the code snippet suggested in the issue https://github.com/ForestAdmin/forest-rails/issues/420.

`next` does not look to be the most readable way to go, it may work to return `next (nil)`, but `return false` is more explicit in `select` block IMO.

I don't get why `!is_active_type?(association.klass)` is not in the `begin`/`rescue` as it is the `association.klass` that looks to be problematic in some cases.

app/services/forest_liana/schema_utils.rb
```diff
@@ -2,9 +2,14 @@ module ForestLiana
...
+        begin
+          association.klass
+        rescue
+          FOREST_LOGGER.warn "Unknown association #{association.name} on class #{active_record_class.name}"
```
arnaudbesnier, line 10: 👍

app/services/forest_liana/schema_utils.rb
```diff
@@ -2,9 +2,14 @@ module ForestLiana
   class SchemaUtils
 
     def self.associations(active_record_class)
-      active_record_class
-        .reflect_on_all_associations
-        .select { |association| !polymorphic?(association) && !is_active_type?(association.klass) }
+      associations = active_record_class.reflect_on_all_associations.select do |association|
+        next if association.options[:polymorphic] && !is_active_type?(association.klass)
+        begin
+          association.klass
```
arnaudbesnier, line 8: it probably works but a boolean is expected here too (implicit return of the select).

arnaudbesnier (changes requested):

arnaudbesnier (approved):

PR #436, opened 2021-03-08
fix(authentication): authentication error when the agent uses a prefix path in its url (#fb79c7)

DrRaider:
See [this blog post](https://www.honeybadger.io/blog/why-is-uri-join-so-counterintuitive/) for the why.

### DIFF
--- a/app/controllers/forest_liana/authentication_controller.rb
+++ b/app/controllers/forest_liana/authentication_controller.rb
@@ -17,7 +17,7 @@ module ForestLiana
     end
   
     def get_callback_url
-        URI.join(ForestLiana.application_url, "/forest/#{CALLBACK_AUTHENTICATION_ROUTE}").to_s
+      File.join(ForestLiana.application_url, "/forest/#{CALLBACK_AUTHENTICATION_ROUTE}").to_s
     rescue => error
       raise "application_url is not valid or not defined" if error.is_a?(ArgumentError)
     end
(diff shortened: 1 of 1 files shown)


ghusse:
It seems that it is not compatible with windows: 

> EDIT 8/12/2016: After publishing this article I received a couple of tweets suggesting I use File.join for this purpose. This has the benefit of avoiding double slashes, ie. /my//path but will break on OSs like Windows, where the path separator isn't a forward-slash.

I think the solution have to be a bit more complex

app/controllers/forest_liana/authentication_controller.rb
```diff
@@ -17,7 +17,7 @@ def initialize
     end
   
     def get_callback_url
-        URI.join(ForestLiana.application_url, "/forest/#{CALLBACK_AUTHENTICATION_ROUTE}").to_s
+      File.join(ForestLiana.application_url, "/forest/#{CALLBACK_AUTHENTICATION_ROUTE}").to_s
```
ghusse, line 20: This should work and be compatible with all OSes

```suggestion
      uri = new URI(ForestLiana.application_url)
      unless ForestLiana.application_url.end_with? '/'
        uri.path += '/'
      uri.path += "forest/#{CALLBACK_AUTHENTICATION_ROUTE}"
      uri.to_str
```

ghusse (changes requested):
Is it possible to add tests on this?

ghusse (approved):
### ✅  &nbsp;Code review

🆗 

### ℹ️   &nbsp;Automatic tests

This specific case is not tested, but it's not simple

### ✅  &nbsp;Manual tests

I could test that this function works as expected, on windows too

### ⚠️&nbsp;PR title

You should consider using a message that expresses more the problem that encountered the client.

Proposition: `fix(authentication): authentication error when the agent uses a prefix path in its url`

### ✅  &nbsp;PR destination branch

🆗

PR #465, opened 2021-05-24
feat(hooks): developers can dynamically add or remove smart actions fields

jeffladiray:
BREAKING CHANGE: fields parameters on hook function is no longer a map of field, it is now an array.
change hook is no longer choosen by the field name, field need to have hook defined inside it definition by addin a props hook.

## Pull Request checklist:

- [ ] Create automatic tests


lib/forest_liana/schema_file_updater.rb
```diff
@@ -96,6 +97,13 @@ def initialize filename, collections, meta
         end
 
         collection['actions'] = collection['actions'].map do |action|
+          begin
+            SmartActionFieldValidator.validate_smart_action_fields(action['fields'], action['name'], action['hooks']['change'])
```
arnaud-moncel, line 101: I think the validation of the hook definition is missing here

app/services/forest_liana/smart_action_field_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+      'File',
+      'Enum',
+    ]
+
+    def self.validate_field(field, action_name)
+      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, nil, "The field attribute must be defined") if !field || field[:field].nil?
+      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, nil, "The field attribute must be a string.") if !field[:field].is_a?(String)
+      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, field[:field], "The description attribute must be a string.") if field[:description] && !field[:description].is_a?(String)
+      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, field[:field], "The enums attribute must be an array.") if field[:enums] && !field[:enums].is_a?(Array)
+      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, field[:field], "The reference attribute must be a string.") if field[:reference] && !field[:reference].is_a?(Array)
```
arnaud-moncel, line 29:
```suggestion
      raise ForestLiana::Errors::SmartActionInvalidFieldError.new(action_name, field[:field], "The reference attribute must be a string.") if field[:reference] && !field[:reference].is_a?(String)
```

arnaud-moncel (changes requested):
There is a little bug but it's amazing 🎉 
Good job 💪

lib/forest_liana/schema_file_updater.rb
jeffladiray: Interesting one. Due to the way you declare the smart action hook I think.
I tend to use `String` instead of Symbol to match the `forest-express-sequelize` contribution (And the existing code in this file), but I wonder what's the best practice here 🤔

arnaud-moncel (approved):
Great job 👌

PR #476, opened 2021-06-01
fix: fix time based graph when timezone is different with database

GuillaumeCisco:
## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/app/services/forest_liana/line_stat_getter.rb
+++ b/app/services/forest_liana/line_stat_getter.rb
@@ -23,7 +23,7 @@ module ForestLiana
     end
 
     def perform
-      value = get_resource().eager_load(@includes)
+      value = get_resource().joins(@includes)
 
       unless @params[:filters].blank?
         value = FiltersParser.new(@params[:filters], value, @params[:timezone]).apply_filters


app/services/forest_liana/line_stat_getter.rb
```diff
@@ -23,7 +23,7 @@ def get_format
     end
 
     def perform
-      value = get_resource().eager_load(@includes)
+      value = get_resource().eager_load(@join)
```
Thenkei, line 26: It's joins if I'm right. https://scoutapm.com/blog/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where
```suggestion
      value = get_resource().eager_load(@joins)
```

Thenkei (commented):
❔ Are you using the right function ?

Thenkei (approved):
### ✅  Code review
It's been a long journey to get this little change. 🔢 
### ⚠️ Automatic tests
We definitely need to add more tests on all this part. Maybe during some Friday afternoon. 
### ✅  Manual tests
We have try every cases. It works as expected. I'm wondering if we will need to go back once the [group date use will be fixed](https://github.com/ankane/groupdate/issues/222). Because `joins` could have lower performances.

PR #479, opened 2021-06-08
fix: allow smart action endpoints to start with slashes

slimee:
Be more flexible about endpoint beginning with slash.

The documentation provides default endpoint starting with slash
/ https://docs.forestadmin.com/documentation/v/v6/reference-guide/actions/create-and-manage-smart-actions#available-smart-action-options

But is was not working when endpoint starts with slash.


GuillaumeCisco (approved):
LGTM 🎉 
Tested together 👯‍♂️

PR #509, opened 2021-07-22
fix: restrict the use of surrounding parentheses only to IN operator

romain-gilliotte:
Although SQLite support the surrounding parentheses for all operators, Postgresql does not.

Note: this branch comes from JoinRaylo:fix-filter-parser-brackets-bug
I'm moving it here so that i can change the commit message, and make it pass the linter step


romain-gilliotte:
closing as the initial PR is getting fixed

PR #522, opened 2021-09-30
fix(authentication): fix certificate issues that break the authentication

arnaudbesnier:
### General

### Security


jeffladiray (approved):

berkos:
Hi @arnaudbesnier @jeffladiray , Thanks very much for the quick response and the resolution of the issue.
Is it worth mentioning in your Readme that you are monkey patching a functionality of a third party gem until [httpclient](https://github.com/nahi/httpclient/issues/445) releases a fix? 🙂

Someone else might be requiring this Gem directly and they would not expect that `forest-liana` would override the same gem too. As a result they could be getting unexpected behaviour.

arnaudbesnier:
Hi @berkos thanks for your suggestion.
What kind of unexpected behaviour do you have in mind that would worth being mentioned in the Readme?

PR #526, opened 2021-10-04
fix: avoid attaching smart action hooks on pre-defined smart actions

jeffladiray:
### General

### Security

### DIFF
--- a/lib/forest_liana/bootstrapper.rb
+++ b/lib/forest_liana/bootstrapper.rb
@@ -57,7 +57,7 @@ module ForestLiana
       @collections_sent.each do |collection|
         collection['actions'].each do |action|
           c = get_collection(collection['name'])
-          unless c.is_nil?
+          unless c.nil?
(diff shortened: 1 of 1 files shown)


lib/forest_liana/bootstrapper.rb
```diff
@@ -57,10 +57,12 @@ def generate_action_hooks()
       @collections_sent.each do |collection|
         collection['actions'].each do |action|
           c = get_collection(collection['name'])
```
Thenkei, line 59:
```suggestion
          `collection` supposed to be the same as `get_collection(collection['name'])` ?
```

Thenkei (approved):
- ✅  Code review - Shouldn't we add a comments ?
- ℹ️   Automatic tests - Not needed, this code is temporary
- ✅  Manual tests
- ❔  PR title - I found the title quite technical but I cannot find better..

(We need to revisit this topic down the road)

PR #566, opened 2022-06-14
fix: prevent duplication of models from multiple ActiveRecord application classes

FraDim:
When the application defines multiple ActiveRecord application classes (for example [`ApplicationRecord`, as Rails documentation currently suggests](https://guides.rubyonrails.org/active_record_basics.html#creating-active-record-models)), the Forest schema contains duplicates of their subclass fields. It seems that the problem stems from a couple of innocent looking code lines:
 - The use of `subclasses` (which only gets direct children) in for the starting point:
 https://github.com/ForestAdmin/forest-rails/blob/d6d381d208335c54056a5dfea0db2281758041cb/lib/forest_liana/bootstrapper.rb#L111 and `descendants` (which gets direct and indirect children) in others:
 https://github.com/ForestAdmin/forest-rails/blob/d6d381d208335c54056a5dfea0db2281758041cb/lib/forest_liana/bootstrapper.rb#L117 and https://github.com/ForestAdmin/forest-rails/blob/d6d381d208335c54056a5dfea0db2281758041cb/lib/forest_liana/bootstrapper.rb#L120

 - The addition of models in the `ForestLiana.models` array when it might already contain them:
 https://github.com/ForestAdmin/forest-rails/blob/d6d381d208335c54056a5dfea0db2281758041cb/lib/forest_liana/bootstrapper.rb#L124
 It seems that some previous code tried to avoid this by using `uniq`:
 https://github.com/ForestAdmin/forest-rails/blob/d6d381d208335c54056a5dfea0db2281758041cb/lib/forest_liana/bootstrapper.rb#L139
 but I think we can prevent those issues (and probably more of the same kind downstream) by allowing only unique models in the array.

I have created two commits:
 - The first one adds and modifies some tests that expose the issue, but does not correct it. This will allow reviewers to see how it manifests.
 - The second one contains my suggested fix.

Please let me know if I should rebase them into one commit.

### General

### Security

### DIFF
--- a/lib/forest_liana/bootstrapper.rb
+++ b/lib/forest_liana/bootstrapper.rb
@@ -114,14 +114,14 @@ module ForestLiana
     def fetch_model(model)
       begin
         if model.abstract_class?
-          model.descendants.each { |submodel| fetch_model(submodel) }
+          model.subclasses.each { |submodel| fetch_model(submodel) }
         else
           if is_sti_parent_model?(model)
-            model.descendants.each { |submodel_sti| fetch_model(submodel_sti) }
+            model.subclasses.each { |submodel_sti| fetch_model(submodel_sti) }
           end
 
           if analyze_model?(model)
-            ForestLiana.models << model
+            ForestLiana.models << model unless ForestLiana.models.include?(model)
           end
         end
       rescue => exception
@@ -136,7 +136,7 @@ module ForestLiana
     end
 
     def create_factories
-      ForestLiana.models.uniq.map do |model|
+      ForestLiana.models.map do |model|
         ForestLiana::SerializerFactory.new.serializer_for(model)
         ForestLiana::ControllerFactory.new.controller_for(model)
       end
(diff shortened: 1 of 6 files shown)


jeffladiray:
Hey @FraDim :wave: and thanks for your contribution on our rails agent 👍

The team will need a bit of time to investigate on both of your contribution (This one and [this one](https://github.com/ForestAdmin/forest-rails/pull/567)), mainly to make sure they match our requirements & that they do not introduce any breaking change.

We will let you know once everything is ready or if any changes are required?

Thanks 🙏

matthv:
Hi @Fradim, thanks for your contribution 💪.

I ran the tests but some of them failed, could you please take a look?

FraDim:
Thanks for informing me @matthv. I have reworked the tests now. Can you please run them again?

matthv (commented):
Sounds good 💪 .
Thank you for your contribution.

PR #567, opened 2022-06-14
fix: handle instance dependent associations

FraDim:
In Rails, when attempting to `eager_load` an association that depends on the instance, I get the following error:

```
/activerecord-7.0.3/lib/active_record/reflection.rb:500:
in `check_eager_loadable!':
The association scope 'eponymous_tree' is instance dependent (the scope block takes an argument).
Eager loading instance dependent scopes is not supported.
(ArgumentError)
```

Since Forest employs `eager_load` to reduce load to the database, an instance dependent association causes errors in table and details views and prevents their use.

The suggested fix allows all Rails version to work. It also contains an optimization of those associations for Rails v7 (the only version that supports this optimization). Older versions of Rails, instead of an error they will work, but will use an unoptimized query.

I have created two commits:
 - The first one adds and modifies some tests that expose the issue, but does not correct it. This will allow reviewers to see how it manifests.
 - The second one contains my suggested fix.

### General

### Security

### DIFF
--- a/app/services/forest_liana/base_getter.rb
+++ b/app/services/forest_liana/base_getter.rb
@@ -29,5 +29,27 @@ module ForestLiana
     def compute_includes
       @includes = ForestLiana::QueryHelper.get_one_association_names_symbol(@resource)
     end
+
+    def optimize_record_loading(resource, records)
+      instance_dependent_associations = instance_dependent_associations(resource)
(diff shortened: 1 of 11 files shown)


matthv:
Hi @FraDim and thanks for your contribution.

I have started the review. You can ignore the test failure on the cover (missing key).

matthv (approved):

PR #569, opened 2022-06-27
fix: take into account subclasses of ActiveRecord models

FraDim:
The previous code, that I changed in https://github.com/ForestAdmin/forest-rails/pull/566, did not account for subclasses of ActiveRecord models. While the case might not happen often, it did change how the gem behaved, so this pull request rectifies this.

I also took the chance to simplify the code that related to single-table-inheritance models (STI). Based on the existing code, both branches of the `if` statement were getting all subclasses no matter if the model was STI or a non-abstract model. This removed the need for the `is_sti_parent_model?` method.

### General

### Security

### DIFF
--- a/lib/forest_liana/bootstrapper.rb
+++ b/lib/forest_liana/bootstrapper.rb
@@ -96,39 +96,25 @@ module ForestLiana
       end
     end
 
-    def is_sti_parent_model?(model)
-      return false unless model.try(:table_exists?)
-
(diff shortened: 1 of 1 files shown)


matthv:
Hi @FraDim, I'm on it  👍.

matthv (approved):

rchasman:
Thank you! Keep up the good work :)

PR #575, opened 2022-07-22
fix: Sort models by name after ActiveRecord::Base.descendants

r4mbo7:
## Definition of Done

Models in `fetch_models` are sorted by alphabetical order as `ActiveRecord::Base.descendants` return value is not deterministic.
So classes order in `.forestadmin-schema.json` is the same for every developers.

(My team have undesirable changes in classes order in `.forestadmin-schema.json` despite using docker).

### General

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [x] Test manually the implemented changes
- [x] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [x] Consider the security impact of the changes made

# Notes

I didn't manage to run spec tests locally !.. This is what I did:
```
docker run -it --rm -v $(pwd):/src/app ruby bash
$ cd /src/app
$ bundle install
$ RAILS_ENV=test bundle exec rake --trace db:migrate test

Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.                                                               
** Invoke load_app (first_time)                                                                                                                                                                                                        
** Execute load_app                                                                                                                                                                                                                    
rake aborted!                                                                                                                                                                                                                          
NameError: uninitialized class variable @@schemes in URI                                                                                                                                                                               
Did you mean?  scheme_list                                                                                                                                                                                                             
/usr/local/bundle/gems/globalid-0.4.2/lib/global_id/uri/gid.rb:176:in `<module:URI>'
...
```
I have tested my changes by editing the gem source file in my project locally and it gives me the expected result.

PR #577, opened 2022-07-22
Use stable sort for schema fields so we get same results across OS

MaicolBen:
We get different results from the schema generation between
Linux and MacOS, the problem is that sort_by isn't stable:
https://stackoverflow.com/a/15442966
https://bugs.ruby-lang.org/issues/1089

The change was introduced recently [here](https://github.com/ForestAdmin/forest-rails/pull/561) and it already has tests (probably hard to test across OS)

### DIFF
--- a/app/services/forest_liana/schema_adapter.rb
+++ b/app/services/forest_liana/schema_adapter.rb
@@ -8,7 +8,7 @@ module ForestLiana
       add_columns
       add_associations
 
-      collection.fields.sort_by! { |k| k[:field].to_s }
+      collection.fields.sort_by!.with_index { |k, idx| [k[:field].to_s, idx] }
 
       # NOTICE: Add ActsAsTaggable fields
       if @model.try(:taggable?) && @model.respond_to?(:acts_as_taggable) &&


r4mbo7:
Thanks @MaicolBen !
Same issue here, other proposal in this PR I've submitted https://github.com/ForestAdmin/forest-rails/pull/575
I hope the FA team will fix this soon :)

MaicolBen:
Oh sorry! I missed it, thank you!

MaicolBen:
@r4mbo7 Your fix doesn't seem to fix the generation in our case, it still generates a different schema for MacOS, so I will keep this open

matthv:
Sorry for my late response.
First of all, I want to be greatful for your contributions @r4mbo7 and @MaicolBen.

As this PR fixes the original issue with a broader compatibility (Both MacOS & Linux), I'll merge this one 👍 

@MaicolBen in the future if you offer another PR, could you use the conventional commit message.

MaicolBen:
Sure! Thank you

PR #572, opened 2022-07-19
fix(multidb): allow associations across databases to work properly

SteveBunlon:
### General

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [x] Test manually the implemented changes
- [x] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [x] Consider the security impact of the changes made

PR #578, opened 2022-07-28
fix(multidb): fix issue with multiple connection detection

lostie:
The previous code in https://github.com/ForestAdmin/forest-rails/pull/572 introduced a couple of issues:

- Some association definitions do not have an `inverse_of` declaration. For those associations the logic was detecting the associations to be running on a different database (which might not be necessarily true). Instead, I used the `klass` method, which Rails always calculates, even for very old versions of Rails. 
- The `resourceDatabase` and `targetModelDatabase` were using different types and therefore all the given `@includes` associations were considered to be running on a different database. I have now used the same method on both, which will correctly detect if they are using the same database.

These were preventing some associations from being eager load correctly and generating SQL.

Before this PR's changes:

```sql
SELECT "products"."id" AS t0_r0,
       "products"."uri" AS t0_r1,
       "products"."name" AS t0_r2,
       "products"."manufacturer_id" AS t0_r3,
       "products"."driver_id" AS t0_r4
FROM "products"
ORDER BY "products"."id" ASC
```

After this PR's changes:

```sql
SELECT "products"."id" AS t0_r0,
       "products"."uri" AS t0_r1,
       "products"."name" AS t0_r2,
       "products"."manufacturer_id" AS t0_r3,
       "products"."driver_id" AS t0_r4,
       "manufacturers"."id" AS t1_r0,
       "manufacturers"."name" AS t1_r1
FROM "products"
     LEFT OUTER JOIN "manufacturers" ON "manufacturers"."id" = "products"."manufacturer_id"
ORDER BY "products"."id" ASC
```
​
I've introduced some test coverage to verify the expected behaviour.

### General

### Security

### DIFF
--- a/spec/services/forest_liana/resources_getter_spec.rb
+++ b/spec/services/forest_liana/resources_getter_spec.rb
@@ -4,9 +4,9 @@ module ForestLiana
     let(:pageSize) { 10 }
     let(:pageNumber) { 1 }
     let(:sort) { 'id' }
-    let(:fields) { }
-    let(:filters) { }
-    let(:scopes) { { } }
+    let(:fields) {}
+    let(:filters) {}
+    let(:scopes) { {} }
     let(:rendering_id) { 13 }
     let(:user) { { 'id' => '1', 'rendering_id' => rendering_id } }
 
@@ -22,7 +22,21 @@ module ForestLiana
       allow(ForestLiana::ScopeManager).to receive(:fetch_scopes).and_return(scopes)
     end
 
+    def clean_database
+      [
+        Driver,
+        Island,
+        Location,
+        Manufacturer,
+        Product,
+        Tree,
+        User,
+      ].each(&:destroy_all)
+    end
+
     before(:each) do
+      clean_database
+
       users = ['Michel', 'Robert', 'Vince', 'Sandro', 'Olesya', 'Romain', 'Valentin', 'Jason', 'Arnaud', 'Jeff', 'Steve', 'Marc', 'Xavier', 'Paul', 'Mickael', 'Mike', 'Maxime', 'Gertrude', 'Monique', 'Mia', 'Rachid', 'Edouard', 'Sacha', 'Caro', 'Amand', 'Nathan', 'Noémie', 'Robin', 'Gaelle', 'Isabelle']
       .map { |name| User.create(name: name) }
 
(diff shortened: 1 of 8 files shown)


spec/services/forest_liana/resources_getter_spec.rb
```diff
@@ -54,11 +68,80 @@ def init_scopes
...
+          connection = resource.connection
+
+          def connection.current_database
+            'db/test.sqlite3'
+          end
+        end
+
+        after do
+          resource.connection.singleton_class.remove_method(:current_database)
+        end
```
SteveBunlon, line 99: ⭐

spec/services/forest_liana/resources_getter_spec.rb
```diff
@@ -54,11 +68,80 @@ def init_scopes
...
+
+        baby_driver = Driver.create!(firstname: 'Baby driver')
+        taxi_driver = Driver.create!(firstname: 'Taxi driver')
+
+        Product.create!(name: 'Valencia', uri: 'https://valencia.com', manufacturer: orange_manufacturer, driver: baby_driver)
+        Product.create!(name: 'Blood', uri: 'https://blood.com', manufacturer: orange_manufacturer, driver: taxi_driver)
+        Product.create!(name: 'Conference', uri: 'https://conference.com', manufacturer: pear_manufacturer, driver: baby_driver)
+        Product.create!(name: 'Concorde', uri: 'https://concorde.com', manufacturer: pear_manufacturer, driver: taxi_driver)
+      end
+
```
SteveBunlon, line 87: Just to follow what has been done previously, I would like this upper in the before each. It's not the best in terms of performances of course, it's just a matter of sticking to what currently exists. 

See the comment upper with the code I suggest ☝️

spec/services/forest_liana/resources_getter_spec.rb
```diff
@@ -54,11 +68,80 @@ def init_scopes
       init_scopes
```
SteveBunlon, line 68:
```suggestion
      manufacturers = [
        { :name => 'Orange' },
        { :name => 'Pear'}
      ].map { |manufacturer| Manufacturer.create(name: manufacturer[:name])}

      drivers = [
        { :firstname => 'Baby driver' },
        { :firstname => 'Taxi driver'}
      ].map { |driver| Driver.create(firstname: driver[:firstname])}

      products = [
        { :name => 'Valencia', :uri => 'https://valencia.com', :manufacturer => manufacturers[0], :driver => drivers[0] },
        { :name => 'Blood', :uri => 'https://blood.com', :manufacturer => manufacturers[0], :driver => drivers[1] },
        { :name => 'Conference', :uri => 'https://conference.com', :manufacturer => manufacturers[1], :driver => drivers[0] },
        { :name => 'Concorde', :uri => 'https://concorde.com', :manufacturer => manufacturers[1], :driver => drivers[1] },
      ].map { |product| Product.create(name: product[:name], uri: product[:uri], manufacturer: product[:manufacturer], driver: product[:driver])}
      
      init_scopes
```

spec/services/forest_liana/resources_getter_spec.rb
SteveBunlon, line 99: I would like to edit this to represent a little bit more what's going on behind the scene. 

```suggestion
      shared_context 'resource current_database' do
        before do
          resource_connection = resource.connection
          target_connection = resource.reflect_on_association(:driver).klass.connection

          def resource_connection.current_database
            'db/test.sqlite3'
          end

          def target_connection.current_database
            'db/test3.sqlite3'
          end
        end

        after do
          resource.connection.singleton_class.remove_method(:current_database)
          resource.reflect_on_association(:driver).klass.connection.singleton_class.remove_method(:current_database)
        end
      end
```

spec/services/forest_liana/resources_getter_spec.rb
lostie: Using index references to records in an array seems a bit more complicated to parse than using named records (I know exactly what the `orange_manufacturer` is, it's hard for me to figure out what `manufacturers[2]`), but if you prefer that format I'll stick to it then.

SteveBunlon: You are totally right, it was just a matter of regrouping records creation for a test suit. I have adapted the code to correspond to your remark 👌

lostie:
@SteveBunlon I can see the `Send coverage` step of the CI is failing but I'm not entirely sure why or how to make it pass. Is this something you have to do on your side?

SteveBunlon (approved):

SteveBunlon:
Hello @lostie 👋 

In fact we have some configuration that prevents the reporter's check to work properly. 

The tests went well, I force merged your contribution. 

Once again, thanks a lot ! Don't hesitate to reach out to us if you encounter anything. Either in the [community](https://community.forestadmin.com/) or here in github. 

Steve.

PR #615, opened 2023-03-22
fix: permission checks using the collection name

lostie:
## Expected outcome

When the app has a model which is defined under a namespace (e.g. `Cars::Motor`)
And the user has permissions to access the Forest collection for that model
And the user is trying to access the data for that Forest collection
Then Forest should render the requested data

## Actual outcome

The Forest agent (`forest_liana`) raises an error resulting into a 500 page in the Forest UI

### General

### Security

### DIFF
--- a/app/services/forest_liana/ability/permission.rb
+++ b/app/services/forest_liana/ability/permission.rb
@@ -8,37 +8,29 @@ module ForestLiana
 
       TTL = (ENV['FOREST_PERMISSIONS_EXPIRATION_IN_SECONDS'] || 1).to_i.second
 
-      def is_crud_authorized?(action, user, collection)
+      def is_crud_authorized?(action, user, model)
         return true unless has_permission_system?
 
-        user_data = get_user_data(user['id'])
-        collections_data = get_collections_permissions_data
+        collection = find_collection!(model)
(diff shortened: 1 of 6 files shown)


lostie:
Hi @matthv , could you please review these changes? Thank you.

lostie:
@nicolasalexandre9 I don't think https://github.com/ForestAdmin/forest-rails/pull/616 fully resolves the issue. Here's my reasoning for it:
- `is_crud_authorized?` was the only method changed in that PR. `is_smart_action_authorized?` has the same problem but it was not changed
- the changes made on the PR look for the name of the collection using `ForestLiana.name_for`. Having tested this method myself, I found that it still returns a valid response even when the provided _collection_ is not actually a ForestAdmin collection:

  ```ruby
  >> ForestLiana.name_for(String)
  
  => "String"
  ```

  so, there's no actual check that the provided _collection_ is a valid ForestAdmin collection.

Apart from the above, the changes made have no test coverage, which makes it more difficult to understand why those changes are being made.

lostie:
Closing, since https://github.com/ForestAdmin/forest-rails/pull/615#issuecomment-1491930509 mentions the issue has been solved.

PR #691, opened 2024-10-15
feat(form): support layout elements in smart action form hooks 

nicolasalexandre9:
### General

### Security

### DIFF
--- a/app/services/forest_liana/smart_action_form_parser.rb
+++ b/app/services/forest_liana/smart_action_form_parser.rb
@@ -7,7 +7,7 @@ module ForestLiana
         if element[:type] == 'Layout'
           validate_layout_element(element)
           element[:component] = element[:component].camelize(:lower)
-          if %w[Page Row].include?(element[:component])
+          if %w[page row].include?(element[:component])
             extract = extract_fields_and_layout_for_component(element)
             layout << element
             fields.concat(extract[:fields])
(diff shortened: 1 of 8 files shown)


app/services/forest_liana/smart_action_form_parser.rb
```diff
@@ -0,0 +1,36 @@
...
+          # TODO: validate layout element
+          if %w[Page Row].include?(element[:component])
+            extract = extract_fields_and_layout_for_component(element)
+            element[:component] = element[:component].camelize(:lower)
+            layout << element
+            fields.concat(extract[:fields])
+          else
+            element[:component] = element[:component].camelize(:lower)
+            layout << element
+          end
```
matthv, line 17: I wonder if by any chance the line
`element[:component] = element[:component].camelize(:lower)`
could be put just above the if? 🤔 

but not sure because of the use of the method: `extract_fields_and_layout_for_component`

nicolasalexandre9: If I put the line before the if the `extract_fields_and_layout_for_component` function will not work.

nicolasalexandre9: I have updated `extract_fields_and_layout_for_component` and add a comment to explain why `Page` becomes `page`

matthv (dismissed):
👍

matthv (approved):
yeah it's better with 100% diff coverage 🎖️

PR #694, opened 2024-10-30
feat(smart-field): add new options to support native polymorphic fields

nicolasalexandre9:
```
field :addressable_type, type: 'String', polymorphic_key: true, is_filterable: true do
  object.addressable_type
end

field :addressable_id, type: 'String', polymorphic_key: true do
  object.addressable_type
end
```

### General

### Security

### DIFF
--- a/lib/forest_liana/collection.rb
+++ b/lib/forest_liana/collection.rb
@@ -72,12 +72,12 @@ module ForestLiana::Collection
       opts[:relationships] = nil unless opts.has_key?(:relationships)
       opts[:widget] = nil unless opts.has_key?(:widget)
       opts[:validations] = [] unless opts.has_key?(:validations)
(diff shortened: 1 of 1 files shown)


matthv (approved):

lostie:
I have tested these changes, but when I define the field for the polymorphic column with `polymorphic_key`, it still require us to provide the options like `is_filterable`, `is_sortable` in order to enable them. 

A nice to have would be if those options were by default true when we mark the field with `polymorphic_key: true`, since those columns are usually filterable and sortable by default anyway.

nicolasalexandre9:
Hi @lostie, thanks for your feedback.

It sounds like a great idea, so we will be opening a PR soon to set these options to true by default in this case.
I'll let you know when the PR is open.

PR #732, opened 2025-05-12
fix: include has_one through associations in compute_select_fields in ResourcesGetter

Pacyfik:
Hi 👋 

## Background
forest_liana 9.12.0 introduced a [db query optimisation](https://github.com/ForestAdmin/forest-rails/pull/729/files) which includes specifying fields which are to be fetched from the database.
This seems to cause an error when requesting an `has_one through` association as one of the fields for the requested resource.

E.g.
Request params:
```
fields: { 'Tree' => 'id,name,location' }
```
```ruby
class Tree < ActiveRecord::Base
  has_one :location, through: :island
end
```
Results in an error:
```
[2025-05-12 16:00:42 +0100] Forest 🌳🌳🌳  Records Index error: missing attribute: island_id
/Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activemodel-6.1.7.9/lib/active_model/attribute.rb:222:in 'ActiveModel::Attribute::Uninitialized#value'
        /Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activemodel-6.1.7.9/lib/active_model/attribute_set/builder.rb:55:in 'block in ActiveModel::LazyAttributeSet#fetch_value'
        /Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activemodel-6.1.7.9/lib/active_model/attribute_set/builder.rb:46:in 'Hash#fetch'
        /Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activemodel-6.1.7.9/lib/active_model/attribute_set/builder.rb:46:in 'ActiveModel::LazyAttributeSet#fetch_value'
        /Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activerecord-6.1.7.9/lib/active_record/attribute_methods/read.rb:30:in 'ActiveRecord::AttributeMethods::Read#read_attribute'
        /Users/mat/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activerecord-6.1.7.9/lib/active_record/attribute_methods.rb:331:in 'ActiveRecord::AttributeMethods#[]'
[...]

PR #742, opened 2025-07-10
fix: include through associations referenced in filters in ResourcesGetter

Pacyfik:
Hi 👋 

## Background
This PR is a follow up to my previous PR: https://github.com/ForestAdmin/forest-rails/pull/732
forest_liana 9.12.0 introduced a [db query optimisation](https://github.com/ForestAdmin/forest-rails/pull/729/files) which includes specifying fields which are to be fetched from the database.
I've found another situation where this causes an issue. Specifically, when there is a filter or multiple filters applied on the forest admin frontend on a through association which is not a displayed column.
I hope the specs I included explain the exact situation. You can reproduce the error if you comment out line 351 of `app/services/forest_liana/resources_getter.rb` (`select << "#{@resource.table_name}.#{association.foreign_key}"`) and run the new specs.

I'd appreciate your opinions on the approach taken and would be happy to help validate any alternative ones.

--

We've also experienced a few similar issues related to polymorphic associations which this PR does not fix.
I've had a hard time investigating those. However, I've found out that uncommenting [this particular line](https://github.com/ForestAdmin/forest-rails/blob/main/lib/forest_liana/active_record_override.rb#L36) in `lib/forest_liana/active_record_override.rb` seems to fix most of them.
Could you consider uncommenting this line in the next release? The issues I've mentioned block us from updating the gem at the moment.

## Affected Versions
Forest agent (forest_liana) >= 9.12.0

### General

### Security

### DIFF
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -347,7 +347,30 @@ module ForestLiana
         end
       end
 
+      associations_from_filters.each do |association|
+        select << "#{@resource.table_name}.#{association.foreign_key}"
+      end
+
       select.uniq
+    end
+
+    def associations_from_filters
+      return [] unless filters_fields
+
+      filters_fields
+        .select { |field| field.include?(':') }
+        .map { |field| field.split(':').first }
+        .map { |field| get_one_association(field) }
+        .compact
+    end
+
+    def filters_fields
+      return unless @params[:filters]
+
+      parsed_filters = JSON.parse(@params[:filters])
+      # Different structure when multiple filters are applied
+      parsed_filters = parsed_filters['conditions'] if parsed_filters['conditions']
+      Array.wrap(parsed_filters).map { |filter| filter['field'] }.compact.uniq
     end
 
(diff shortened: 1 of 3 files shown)


matthv:
I spent some time on this today and noticed that the `extract_associations_from_filter` method wasn't behaving correctly. I took the opportunity to patch it in [this draft branch](https://github.com/ForestAdmin/forest-rails/tree/fix/properly-extract-associations-from-filters), which also incorporates some of the logic you proposed in your patch.

Could you let me know if this fixes the issue you’ve been encountering?

I'm also going to test your branch and will get back to you shortly.
Finally, I'm looking into the other issue you mentioned regarding polymorphic associations as well.

Thanks again for all the detailed feedback! 🙏

matthv:
I’ve just merged the first fix v9.15.1. I’ll take care of the other issue as soon as I have some time.

In the meantime, I’ve created another [branch](https://github.com/ForestAdmin/forest-rails/tree/fix/polymorphic) with the line uncommented to unblock you for now.

PR #759, opened 2025-10-09
feat: add feature flag to control usage of query optimizations

lostie:
## Context

I've encountered several issues with the query optimizations:
- the SQL produced for retrieving records with `has_one` associations is invalid (it's currently trying to select the foreign key from the main record's table) resulting in errors when rendering the view or exporting the view to CSV
- the SQL produced when the record has composite foreign keys is invalid resulting in errors when rendering the view

I've attempted to fix some of these issues myself, but I ended up realizing that the gem code needs extensive refactoring (and added test coverage) and unfortunately I don't have the necessary time to spend on it. 

These query optimizations are **breaking changes**, and have prevented us from upgrading the gem since the version `9.11.3`, so my suggestion is to allows us to use the gem without the query optimizations (which we know it is at least functional).

### General

### Security

### DIFF
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -30,17 +30,19 @@ module ForestLiana
     end
 
     def perform
-      polymorphic_association, preload_loads = analyze_associations(@resource)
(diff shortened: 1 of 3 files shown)


lostie:
Closing since we decided to take a different approach to address the issue.

PR #770, opened 2026-05-21
fix: search array uuid columns using ANY() operator

ramongtx:
## Problem

When performing a full-text search on a collection that contains `uuid[]` (array of UUID) columns, Forest Rails generates a plain `=` comparison:

```sql
"attachment_ids" = '018e7fc6-d8a6-7ceb-b80e-d0972603e1b5'
```

PostgreSQL rejects this with:

```
PG::InvalidTextRepresentation: ERROR: malformed array literal: "018e7fc6-..."
DETAIL: Array value must start with "{" or dimension information.
```

This causes a 500 error and locks agents from performing searches on any collection that has array-typed UUID fields.

## Solution

When the search term is a valid UUID and the target column is of type `:uuid`, check whether it is also an array column. If so, use the PostgreSQL `ANY()` operator instead of `=`:

```sql
'018e7fc6-d8a6-7ceb-b80e-d0972603e1b5' = ANY("attachment_ids")
```

This correctly checks whether the UUID appears anywhere in the array.

Non-array UUID columns and all other column types are unaffected. The `column.array` attribute is PostgreSQL-specific, so MySQL and SQLite are also unaffected.

## Changes

- `app/services/forest_liana/search_query_builder.rb`: branch on `column.array` when building the UUID search condition
- `spec/services/forest_liana/search_query_builder_spec.rb`: new spec covering both the `ANY()` path for array columns and the `=` path for scalar columns

> [!NOTE]
> ### Fix UUID search on array columns using `ANY()` operator
> Equality checks on PostgreSQL `uuid[]` array columns always failed when searching by UUID. [search_query_builder.rb](https://github.com/ForestAdmin/forest-rails/pull/770/files#diff-fa135f02c0c5311220e1b811b2332b4c7e337954342efc5a94503d847d836841) now detects array-typed UUID columns and switches to a `:search_value_for_uuid = ANY(column_name)` condition, while non-array UUID columns retain the existing equality check.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ebabca5.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -75,7 +75,11 @@ module ForestLiana
             end
           # NOTICE: Rails 3 do not have a defined_enums method
           elsif REGEX_UUID.match(@search) && column.type == :uuid
-            conditions << "#{column_name}  = :search_value_for_uuid"
+            if column.respond_to?(:array) && column.array
+              conditions << ":search_value_for_uuid = ANY(#{column_name})"
+            else
+              conditions << "#{column_name}  = :search_value_for_uuid"
+            end
           elsif @resource.respond_to?(:defined_enums) &&
             @resource.defined_enums.has_key?(column.name) &&
             !@resource.defined_enums[column.name][@search.downcase].nil?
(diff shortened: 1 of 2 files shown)


matthv (approved):
Thanks for catching this and for the clean patch! LGTM

PR #772, opened 2026-06-03
fix(schema): emit is_primary_key flag on collection fields

christophebrun-forest:
## Problem

Forest Admin workflows in **Orchestrator** mode (run by the workflow-executor) fail on a simple read-record with:

> Collection "Author" has no primary key field in its rendering layout. Workflow runs cannot operate on collections without an addressable primary key.

**Browser** mode works, Orchestrator does not.

## Root cause

The SaaS (`workflow-orchestrator/collection-schema`) derives the primary key from the `is_primary_key` flag in the schema pushed by the agent. The gem **never emits** `is_primary_key`, so the SaaS finds no addressable PK for **any** collection of a Ruby agent.

## Fix

Emit `is_primary_key` (boolean) on every column field, `true` only for the column(s) composing the ActiveRecord primary key (handles composite PKs via `composite_primary_keys` where `primary_key` returns an `Array`, and the `nil` case). Default `false` everywhere else, associations included.

### 5 places kept consistent
- **`schema_adapter.rb`** — computed in `get_schema_for_column`; `false` on every association path (`get_schema_for_association`, the inline polymorphic hash, and the FK→association reuse path in `add_associations`).
- **`apimap_sorter.rb`** — whitelisted in `KEYS_COLLECTION_FIELD`. The schema sent to the SaaS goes through `send_apimap → SchemaSerializer → ApimapSorter#reorder_collection_fields → POST /forest/apimaps`, which `slice`s by this list; an unlisted key is **stripped** in both dev and prod.
- **`lib/forest_liana/collection.rb`** (the `field` DSL) — default `false` for scalar smart fields (these are defaulted here, not in `init_properties_with_default`).
- **`schema_file_updater.rb`** — whitelist + default `false` for the `.forestadmin-schema.json` file.
- **`model/collection.rb`** — default `false` for the `Collection.new(fields:)` path.

## Tests
- New minitest cases: standard PK column → `true`, non-PK column → `false`, composite PK → `true` on each PK column, association → `false`. Assertions on the exact flag value; tests call `get_schema_for_column`/`get_schema_for_association` directly to avoid polluting the shared global apimap.
- Updated snapshot specs (`schema_adapter_spec` addressable field, `collection_spec` smart field).
- Full suite green locally: `rake test` 19/0, `rspec` 419/0.

## Notes
- Scope strictly limited to the `is_primary_key` flag.
- A pre-existing `Gemfile.lock` inconsistency on `main` was noticed (`Gemfile` pins `minitest ~> 5.25` but the lock has `6.0.1`); **not** touched here — worth a separate change.

Ref: PRD-439

> [!NOTE]
> ### Emit `is_primary_key` flag on all collection fields in the schema
> - Adds `is_primary_key` to every field hash in the generated schema: `true` for columns that are part of the model's primary key (including composite keys), `false` for all other columns and association fields.
> - Introduces a `primary_key?` helper in [`SchemaAdapter`](https://github.com/ForestAdmin/forest-rails/pull/772/files#diff-1eae726941328a7d207dbdfd5b67365d6c4f37e62752bdbf68e8d4f1e2a4b4cf) that handles both singular and array primary key values.
> - Updates [`ApimapSorter`](https://github.com/ForestAdmin/forest-rails/pull/772/files#diff-177c72307171d214d6643440839f8baadc48cc38dbb7d59ae5b19714f18ebd6b) and [`SchemaFileUpdater`](https://github.com/ForestAdmin/forest-rails/pull/772/files#diff-751b9fd2ad74012cdd14b46e1779db0e2339f03161f642da3f093a779548b9fa) to include `is_primary_key` in their field key whitelists so the flag is retained and written to the schema file.
> - Smart fields defined via `ForestLiana::Collection.field` also default to `is_primary_key: false`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 025a424.</sup>
> 
>

### DIFF
--- a/app/models/forest_liana/model/collection.rb
+++ b/app/models/forest_liana/model/collection.rb
@@ -37,6 +37,7 @@ class ForestLiana::Model::Collection
       field[:enums] = nil unless field.key?(:enums)
       field[:integration] = nil unless field.key?(:integration)
       field[:is_filterable] = true unless field.key?(:is_filterable)
+      field[:is_primary_key] = false unless field.key?(:is_primary_key)
       field[:is_read_only] = false unless field.key?(:is_read_only)
       field[:is_required] = false unless field.key?(:is_required)
       field[:is_sortable] = true unless field.key?(:is_sortable)
--- a/app/services/forest_liana/apimap_sorter.rb
+++ b/app/services/forest_liana/apimap_sorter.rb
@@ -20,6 +20,7 @@ module ForestLiana
       'enums',
       'integration',
       'is_filterable',
+      'is_primary_key',
       'is_read_only',
       'is_required',
       'is_sortable',
--- a/app/services/forest_liana/schema_adapter.rb
+++ b/app/services/forest_liana/schema_adapter.rb
@@ -248,6 +248,7 @@ module ForestLiana
               relationship: get_relationship_type(association),
               reference: "#{association.name.to_s}.id",
               inverse_of: @model.name.demodulize.underscore,
+              is_primary_key: false,
               is_filterable: false,
               is_sortable: true,
               is_read_only: false,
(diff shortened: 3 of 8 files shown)


matthv (approved):
LGTM

PR #774, opened 2026-06-04
fix(composite-pk): accept pipe-joined composite ids [PRD-450]

matthv:
## Problem
The workflow executor (via `@forestadmin/agent-client`) addresses records by a **pipe-joined** composite id (`k1|k2`), but `parse_composite_id` only accepted the **JSON** `[k1,k2]` form the frontend sends and raised `HTTP422` otherwise. As a result, any composite-key **update / by-id** coming from the executor failed with a 422.

## Fix
`parse_composite_id` now accepts **both** shapes:
- JSON array `"[k1,k2]"` — from the Forest frontend
- pipe-joined `"k1|k2"` — from the agent-client (workflow executor, Forest convention)

Both keep the values in `primary_key` order, so `find_record`'s `zip` against `resource.primary_key` is unchanged and the frontend keeps working as before. No serializer/output change.

## Scope
This is the **agent (Ruby) part** of PRD-450. The executor-side fix (reading composite records via the by-id route + related-data) lands in a separate `agent-nodejs` PR.

## Testing
- `composite_primary_keys_spec.rb`: added a pipe-parsing test (the obsolete "raises HTTP422 on invalid format" case is replaced).
- Validated end-to-end: a composite-key `Shipping` workflow (get-data → condition → update-data) now **reads and updates** the record in orchestrator mode — no more `Record not found` / 422.

Relates to PRD-450.

> [!NOTE]
> ### Accept pipe-joined composite IDs in `RecordFindable.parse_composite_id`
> Previously, `parse_composite_id` only accepted JSON array strings and raised `HTTP422Error` for any other format. It now also splits on `|` for pipe-joined IDs (e.g. `"v1|v2"`), returning an array of string values. The spec in [composite_primary_keys_spec.rb](https://github.com/ForestAdmin/forest-rails/pull/774/files#diff-fc1f76b54ca665ba43044d12c8b28e0405a58c8da95407cf2705008c2c3e0f9b) is updated to cover both formats and removes the test expecting an error on non-JSON input.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 120ea57. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/app/services/forest_liana/record_findable.rb
+++ b/app/services/forest_liana/record_findable.rb
@@ -14,13 +14,18 @@ module ForestLiana
       end
     end
 
+    # Composite ids reach us in two shapes, both ordered like the model's primary_key:
+    #   - JSON array "[v1,v2]"  -> from the Forest frontend
+    #   - pipe-joined "v1|v2"   -> from the agent-client (workflow executor, Forest convention)
+    # Either way the values stay in primary_key order, so find_record can zip them directly.
     def parse_composite_id(id)
       return id if id.is_a?(Array)
 
-      if id.to_s.start_with?('[') && id.to_s.end_with?(']')
-        JSON.parse(id.to_s)
+      str = id.to_s
+      if str.start_with?('[') && str.end_with?(']')
+        JSON.parse(str)
       else
-        raise ForestLiana::Errors::HTTP422Error.new("Composite primary key ID must be in format [value1,value2], received: #{id}")
+        str.split('|')
       end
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
ok for me

PR #777, opened 2026-06-22
refactor(workflow): forward workflow executor routes generically [PRD-567]

matthv:
## What

Phase 3 (Ruby v1 liana) of PRD-567. Replaces the two explicit `show`/`trigger` proxy routes with a **single catch-all** that forwards every verb and sub-path under `/_internal/workflow-executions/*` to the executor `/runs` namespace. The agent no longer needs a change when the executor adds a route.

Mirrors the Node ([agent-nodejs#1666](https://github.com/ForestAdmin/agent-nodejs/pull/1666)) and Ruby v2 ([agent-ruby#319](https://github.com/ForestAdmin/agent-ruby/pull/319)) implementations.

## Changes

- **Catch-all glob route** (`match '*path', via: :all`) replacing the 2 per-route entries — all verbs.
- **Forward all request headers** except hop-by-hop / host / content-length (was an `Authorization`+`Cookie` whitelist).
- **Forward all response headers** except hop-by-hop — so executor-set headers (e.g. the version gate) survive the proxy instead of being dropped.
- **Raw request body** forwarded untouched.
- **Single large timeout** (open 2s / read 120s), no per-route values.
- **Path-traversal guard**: the wildcard can only map into `/runs` — `..`, encoded dots, backslash, null byte and leading `/` are rejected with 404, keeping non-`/runs` executor routes (e.g. mcp-oauth-credentials) unreachable through the proxy.

## Tests

Rewritten request spec → generic forwarding (arbitrary verb/sub-path), raw body, request + response header filtering, **path-traversal security boundary**, status/error passthrough, 401 unauthenticated, 404 when unconfigured. **14/14 green** under Ruby 3.3.

fixes PRD-567

> [!NOTE]
> ### Replace specific workflow executor routes with a generic catch-all proxy for all verbs under `/runs/*`
> - Replaces two fixed routes (`GET /runs/:run_id` and `POST /runs/:run_id/trigger`) with a single catch-all route matching all HTTP verbs under `/_internal/workflow-executions/*path` in [routes.rb](https://github.com/ForestAdmin/forest-rails/pull/777/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5).
> - The new `proxy` action in [WorkflowExecutionsController](https://github.com/ForestAdmin/forest-rails/pull/777/files#diff-c4d799f239b33ca99911c5d701d11408d1df0cb69e5a9a8723fdb964d2ec83e0) forwards any verb and sub-path to the executor, including query params, raw body, and request headers (excluding hop-by-hop, Host, and encoding headers).
> - Upstream response headers are now propagated back to the client, and SSL/transport errors return 503.
> - Path traversal attempts (empty, absolute, or `..`-containing paths) are rejected with 404 via `safe_executor_path`.
> - Behavioral Change: open timeout is 2s and overall request timeout is 120s; `OpenSSL::SSL::SSLError` is now caught and returned as 503.
>
> #### 🖇️ Linked Issues
> Relates to [PRD-567](ticket:linear/PRD-567).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1644bb0.</sup>
> 
>

## Known limitation (by design)

This legacy liana is **JSON-only**: responses are rendered as JSON (`render json: ...`). A hypothetical non-JSON executor response (e.g. `text/plain`) would not pass through verbatim. The executor only returns JSON today, so this does not affect the PRD-567 contract in practice; the `@forestadmin/agent` Node core ([agent-nodejs#1666](https://github.com/ForestAdmin/agent-nodejs/pull/1666)) handles raw passthrough.

### DIFF
--- a/app/controllers/forest_liana/workflow_executions_controller.rb
+++ b/app/controllers/forest_liana/workflow_executions_controller.rb
@@ -2,8 +2,18 @@ require 'httparty'
 
 module ForestLiana
   class WorkflowExecutionsController < ApplicationController
-    FORWARDED_HEADERS = %w[Authorization Cookie].freeze
-    UPSTREAM_TIMEOUT_IN_SECONDS = 10
+    EXECUTOR_PREFIX = '/runs'.freeze
+    # Hop-by-hop headers + those the HTTP client / framework recomputes — never forwarded
+    # (request or response side).
+    SKIPPED_HEADERS = %w[
+      connection keep-alive transfer-encoding upgrade te trailer
+      proxy-authenticate proxy-authorization host content-length
+    ].freeze
+    # Substrings that could let the wildcard escape EXECUTOR_PREFIX (traversal, encoded dots,
+    # backslash, null byte).
+    UNSAFE_PATH_FRAGMENTS = ['..', '%2e', '%2E', '\\', "\0"].freeze
+    OPEN_TIMEOUT_IN_SECONDS = 2
+    REQUEST_TIMEOUT_IN_SECONDS = 120
     UPSTREAM_ERRORS = [
       HTTParty::Error,
       SocketError,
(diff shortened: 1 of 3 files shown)


christophebrun-forest (approved):
Ok for me

PR #778, opened 2026-06-25
feat(workflow): forward any executor route generically [PRD-567]

matthv:
## What

Ports the generic executor passthrough to forest-rails (Ruby v1 liana), matching the agent-nodejs reference ([agent-nodejs#1707](https://github.com/ForestAdmin/agent-nodejs/pull/1707), 1.81.0).

- **Single catch-all mount** `_internal/executor/*path` → executor `/:path` **verbatim** (all verbs). Replaces the `/runs`-confined proxy. Callers address runs as `runs/<id>`.

## Security

- **SSRF guard**: string guard (empty / leading-`/` / `..` / `%2e`/`%2E` / backslash / NUL → 404) plus a `URI`-based same-origin check (reject unless scheme+host+port match; a control char like `\t//host` makes `URI.parse` raise → 404).
- **`follow_redirects: false`**: HTTParty follows redirects by default — a 3xx `Location` to an off-origin host would bypass the origin guard, so disable it (the Node agent's raw http client never follows redirects).

## Tests

**17/17**. Run GET (caller includes `runs/`), POST trigger, **non-runs route verbatim** (no `/runs` prefix injected), `follow_redirects: false` asserted, header req/resp filtering, status/body passthrough, SSRF vectors → 404, 401 unauthenticated, unreachable/SSL → 503, 404 when unconfigured.

fixes PRD-567

> [!NOTE]
> ### Refactor `WorkflowExecutionsController` proxy to forward executor routes generically with same-origin enforcement
> - Renames the proxy route from `_internal/workflow-executions/*path` to `_internal/executor/*path` in [config/routes.rb](https://github.com/ForestAdmin/forest-rails/pull/778/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5).
> - Removes the automatic `/runs` prefix injection; the proxy now forwards the client-provided sub-path verbatim via `safe_executor_path`.
> - Adds a `same_origin?` helper that parses scheme, host, and port to reject any forwarding target that differs from the configured executor base URL.
> - Sets `follow_redirects: false` in HTTParty to prevent off-origin redirect exploitation.
> - Behavioral Change: any existing clients calling `/forest/_internal/workflow-executions/...` will receive 404; callers must update to `/forest/_internal/executor/...`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2d7263f.</sup>
> 
>

### DIFF
--- a/app/controllers/forest_liana/workflow_executions_controller.rb
+++ b/app/controllers/forest_liana/workflow_executions_controller.rb
@@ -1,8 +1,8 @@
 require 'httparty'
+require 'uri'
 
 module ForestLiana
   class WorkflowExecutionsController < ApplicationController
-    EXECUTOR_PREFIX = '/runs'.freeze
     # Never forwarded (request or response): hop-by-hop, Host, and body-framing headers.
     # render json: re-serializes the body, so the upstream length/encoding no longer match — and
     # forwarding accept-encoding would defeat Net::HTTP's transparent gzip decompression.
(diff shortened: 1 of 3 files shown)


app/controllers/forest_liana/workflow_executions_controller.rb
```diff
@@ -53,14 +57,22 @@ def proxy
...
+    # First-pass rejection of escape attempts; the authoritative origin check is same_origin?.
     def safe_executor_path
       wildcard = params[:path].to_s
       return nil if wildcard.empty? || wildcard.start_with?('/')
       return nil if UNSAFE_PATH_FRAGMENTS.any? { |fragment| wildcard.include?(fragment) }
 
-      "#{EXECUTOR_PREFIX}/#{wildcard}"
+      "/#{wildcard}"
+    end
+
```
christophebrun-forest, line 68: OK, but could be merged. Two methods with some redundancies regarding SSRF.

christophebrun-forest (approved):
ok, just small redondancies

app/controllers/forest_liana/workflow_executions_controller.rb
matthv: Intentional split: safe_executor_path rejects path traversal (.. etc.) before we even call the executor, while same_origin? is the actual guard that ensures we never leave the executor origin. I kept them separate to match the other agents, so the SSRF model reads the same everywhere.

PR #782, opened 2026-07-30
perf: avoid a SQL query per association in separate_database?

Slpn:
## What

`BaseGetter#separate_database?` decides whether an association can be eager-loaded via a JOIN (same database) or must be preloaded separately (cross-database). It currently calls `connection.current_database` — a `SELECT current_database()` round-trip — twice per association (resource + association connection), on every request.

The database a connection points to is fixed at boot, so this reads it from the in-memory pool config instead:

```ruby
association.klass.connection.pool.db_config.database
resource.connection.pool.db_config.database
```

On apps with many collections across multiple databases, this removes a large share of read-replica queries.

## Why it's safe

- `connection.pool.db_config` is the same source `SchemaUtils#disable_filter_and_sort_if_cross_db!` already uses for cross-database detection.
- It keeps going through `.connection`, so it stays correct whether `resource` is a class (`ResourcesGetter`), a record instance, or a `CollectionProxy` (`HasManyGetter`). Reading `db_config` off the class-only `connection_db_config` would break the `HasManyGetter` path.
- The `respond_to?(:current_database)` guard is dropped — every adapter has `pool.db_config`. Bonus: two different SQLite databases are now correctly detected as separate (previously both returned `nil`, so they were treated as the same).

## Tests

Reworked the specs to use the dummy app's real multi-database setup (`Product` on `test`, `Driver` on `user`) instead of stubbing `current_database`, plus a test asserting `separate_database?` issues no SQL.

> [!NOTE]
> ### Avoid SQL queries in `separate_database?` by using `db_config` instead of `current_database`
> - Replaces `connection.current_database` with `connection.pool.db_config.database` in [`separate_database?`](https://github.com/ForestAdmin/forest-rails/pull/782/files#diff-b8f760393646995fca05cc29342760c53c6e7bb3a6544cedc7d351b2b8ebf315) to determine whether an association lives in a different database.
> - This avoids issuing a SQL query per association when checking database separation, and also handles cases where `current_database` is undefined.
> - Test helpers that monkey-patched `current_database` are removed and replaced with a spec that asserts no SQL queries are triggered by `separate_database?`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 72ad471.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/base_getter.rb
+++ b/app/services/forest_liana/base_getter.rb
@@ -51,10 +51,8 @@ module ForestLiana
     def separate_database?(resource, association)
       return false if SchemaUtils.polymorphic?(association)
 
-      target_model_connection = association.klass.connection
-      target_model_database = target_model_connection.current_database if target_model_connection.respond_to? :current_database
-      resource_connection = resource.connection
-      resource_database = resource_connection.current_database if resource_connection.respond_to? :current_database
+      target_model_database = association.klass.connection.pool.db_config.database
+      resource_database = resource.connection.pool.db_config.database
 
       target_model_database != resource_database
     end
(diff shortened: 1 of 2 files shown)


matthv:
Hello @Slpn,
Thanks for the contribution 🙏

PR #784, opened 2026-07-31
fix: translate null equality/inequality filters to IS [NOT] NULL

ramongtx:
## Problem

When a user sets an equality filter to null, the Forest web app sends it as a
**JSON null** — e.g. an unresolved `{{currentUser.tags.…}}` template — which
reaches Ruby as `nil`. `FiltersParser` then builds `field = NULL` /
`field != NULL`, and neither ever matches anything in SQL, so the filter
silently returns zero rows.

(An exact-text search for the literal string `"null"` is intentionally left
alone, so a search for a row whose value is the text `"null"` keeps working.)

## Fix

Key the guard on `value.nil?`, captured **before** the enum mapping (which nils
unknown enum values and would otherwise look like a null filter), and cover both
operators:

```rb
null_value = value.nil?

# ... later, right after parsed_field ...
if null_value
  return "#{parsed_field} IS NULL" if operator == 'equal'
  return "#{parsed_field} IS NOT NULL" if operator == 'not_equal'
end
```

## Test

In `spec/services/forest_liana/filters_parser_spec.rb`:
- `equal` on a JSON null asserts `pluck(:name)` matches the NULL row and leaves a
  seeded `Tree.create!(name: 'null')` row unmatched (pins string-vs-null).
- `not_equal` on a JSON null matches the non-null rows (`IS NOT NULL`).
- a regression pinning the capture-before-enum ordering (`cutter:title equal
  "<unknown enum>"` stays `= NULL` → 0 rows).

95 examples / 0 failures; the two null specs go red when only the source file is
reverted. Full suite matches `main`.

### General

### Security

- [x] Consider the security impact of the changes made — the guard emits a fixed
  `IS [NOT] NULL` SQL fragment (no interpolation of user input).

### DIFF
--- a/app/services/forest_liana/filters_parser.rb
+++ b/app/services/forest_liana/filters_parser.rb
@@ -83,6 +83,10 @@ module ForestLiana
       value = condition['value']
       field_name = condition['field']
 
+      # NOTICE: captured before the enum mapping below, which turns an unknown
+      #         enum value into nil and would otherwise look like a null filter.
+      null_value = value.nil?
+
       if @operator_date_parser.is_date_operator?(operator)
         field_schema = SchemaUtils.find_column_schema_by_name(ForestLiana.name_for(@resource), field_name)
         condition = @operator_date_parser.get_date_filter(operator, value, field_schema)
@@ -99,6 +103,14 @@ module ForestLiana
 
       case_insensitive = operator == 'i_contains'
       parsed_field = parse_field_name(field_name, case_insensitive)
+
+      # NOTICE: the UI sends a null equality filter as a JSON null; `= NULL` and
+      #         `!= NULL` never match anything in SQL.
+      if null_value
+        return "#{parsed_field} IS NULL" if operator == 'equal'
+        return "#{parsed_field} IS NOT NULL" if operator == 'not_equal'
+      end
+
       parsed_operator = parse_operator(operator)
       parsed_value = parse_value(operator, value)
       field_and_operator = "#{parsed_field} #{parsed_operator}"
--- a/spec/services/forest_liana/filters_parser_spec.rb
+++ b/spec/services/forest_liana/filters_parser_spec.rb
@@ -45,6 +45,29 @@ module ForestLiana
             it { expect(parsed_filters.count).to eq 1 }
           end
 
+          context 'equal on a null value (JSON null from the UI)' do
+            before do
+              Tree.create!(name: nil, age: 42, island: Island.first)
+              Tree.create!(name: 'null', age: 43, island: Island.first)
+            end
+
+            let(:filters) { { 'field' => 'name', 'operator' => 'equal', 'value' => nil } }
+
+            it 'matches rows where the column IS NULL, not the literal string "null"' do
+              expect(parsed_filters.pluck(:name)).to eq([nil])
+            end
+          end
+
+          context 'not_equal on a null value (JSON null from the UI)' do
+            before { Tree.create!(name: nil, age: 42, island: Island.first) }
+
+            let(:filters) { { 'field' => 'name', 'operator' => 'not_equal', 'value' => nil } }
+
+            it 'matches rows where the column IS NOT NULL' do
+              expect(parsed_filters.pluck(:name)).to match_array(['Tree n1', 'Tree n2', 'Tree n3'])
+            end
+          end
+
           context 'greater_than' do
             let(:filters) { { 'field' => 'age', 'operator' => 'greater_than', 'value' => 2 } }
             it { expect(parsed_filters.count).to eq 2 }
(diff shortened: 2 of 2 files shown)


bexchauveto (changes requested):
Thanks for the repro and for shipping a test with it. I dug into this one and I think the guard is keyed on the wrong value — details below, with what I ran.

## What the UI actually sends

I ran a unit test against `fragment-condition`'s `queryCondition` in the frontend repo, with a real `currentUser` carrying one tag:

| user configures | `isValidData` | payload |
|---|---|---|
| plain text `hello` | true | `{"operator":"equal","value":"hello"}` |
| **types the 4 chars `null`** | true | `{"operator":"equal","value":"null"}` ← String |
| `{{currentUser.tags.existing-tag}}` | true | `{"operator":"equal","value":"hello"}` |
| **`{{currentUser.tags.missing-tag}}`** | true | `{"operator":"equal","value":null}` ← JSON null |
| empty value | **false** | condition dropped, never sent |

Two things fall out of that:

**1. The reported bug comes from the JSON-null payload, which this PR doesn't touch.** An unresolved templating key — a `currentUser` tag absent on the requesting user — serializes to `"value":null` (the `?? null` in `fragment-condition.ts:457`; `isValidData` and `isContextSet` are both true, so the condition isn't dropped). Filters are `JSON.parse`d at `resources_getter.rb:157`, so that arrives as Ruby `nil`, `value == 'null'` is false, and we still emit `WHERE name = NULL` → zero rows. I confirmed via `GET /forest/Tree?filters=…`: identical empty result on `main` and on this branch.

**2. The string `"null"` only arrives when a user literally types it** — i.e. an exact-text search, which is the one case where `IS NULL` is the wrong answer. Verified against three rows (`"Tree n1"`, `nil`, `"null"`): on `main` that filter returns the `"null"` row; on this branch it returns the `nil` row, and the `"null"` row becomes unreachable by exact match. Also worth noting the UI already ships `blank` for null checks (`common-operators.ts:99-104`), which emits byte-identical SQL — `"trees"."name" IS NULL`.

## Suggested change

Same one line, keyed on nil, plus the `not_equal` counterpart (the templating path hits it identically — `!= NULL` matches nothing where the user means `IS NOT NULL`):

```rb
operator = condition['operator']
value = condition['value']
field_name = condition['field']

# NOTICE: captured before the enum mapping below, which turns any unknown
#         enum value into nil and would otherwise look like a null filter.
null_value = value.nil?

# ... later, right after parsed_field ...

# NOTICE: the UI sends a null equality filter as a JSON null; `= NULL` and
#         `!= NULL` never match anything in SQL.
if null_value
  return "#{parsed_field} IS NULL" if operator == 'equal'
  return "#{parsed_field} IS NOT NULL" if operator == 'not_equal'
end
```

The nil capture has to happen **before** the enum mapping at lines 96-98, which nils out unknown enum values. I built both orderings to check: with the check placed after, `title equal "bogus_enum"` wrongly returns every null-title row; placed before, it stays `= NULL` → no rows, matching `main`.

## What I verified

- HTTP probe through `GET /forest/Tree?filters=…` on `main`, this branch, and the `value.nil?` variant — `equal`/`not_equal`, JSON null vs string, `and` aggregators, belongs-to fields, integer and enum columns.
- `filters_parser_spec.rb`: 96 examples, 0 failures with the change; the two new null specs go red when only the source file is reverted.
- Full suite: failure set byte-identical to `main` for both this PR and the variant — neither introduces regressions.

## On the spec

`expect(parsed_filters.count).to eq 1` passes because exactly one tree has a nil name — it never asserts *which* row matched, so it would pass for the wrong reason. Worth asserting `pluck(:name)` and adding a `Tree.create!(name: 'null')` row, so the string-vs-null distinction is pinned rather than incidental.

Nit: the 4-line `# NOTICE:` is heavier than the one-line guard needs — only "a plain `= NULL` never matches" is non-obvious. And the spec comment restates the `context` string directly above it.

Happy to push my branch as an alternative if that's easier than reworking this one.

bexchauveto (approved):
Code and specs look good — thanks for the quick turnaround.

One thing left before merge: the PR title and description still describe the previous approach.

- **Title** is still `fix: translate equal filter with literal "null" to IS NULL`, but the guard no longer keys on the literal string — it keys on a JSON null, and it now covers `not_equal` too. The commit message (`fix: translate null equality/inequality filters to IS [NOT] NULL`) is already correct; the title should match it.
- **Description** still states that the web app "sends the value as the literal string `null`", and the `## Fix` block still shows `value == 'null'`. Both describe behaviour the branch no longer has — the string case is now left alone on purpose, so that an exact-text search for `null` keeps working.

Worth fixing because this repo runs semantic-release: if the merge takes the PR title, the changelog entry will describe the wrong behaviour.

PR #785, opened 2026-07-31
fix: map timestamptz columns to the Date type

ramongtx:
## Problem

`SchemaAdapter#get_type_for` maps `:datetime` columns to the Forest `Date` type,
but PostgreSQL `timestamptz` columns are reported by ActiveRecord with the type
`:timestamptz`, which is not covered by the `case` statement. Such columns fall
through to a `nil` type and are silently dropped from the generated schema.

## Fix

Map `:timestamptz` alongside `:datetime`:

```rb
when :datetime, :timestamptz
  type = 'Date'
```

(`:citext` and array columns are already handled upstream, so this is the only
remaining gap.)

## Test

Added unit specs in `spec/services/forest_liana/schema_adapter_spec.rb` asserting
`get_type_for` maps both `:timestamptz` and `:datetime` columns to `Date`.
Verified red before / green after. Full suite green.

### General

### Security

- [x] Consider the security impact of the changes made — none (static type mapping).

> [!NOTE]
> ### Map `:timestamptz` and `:timestamp` column types to 'Date' in `SchemaAdapter`
> Extends `ForestLiana::SchemaAdapter.get_type_for` in [schema_adapter.rb](https://github.com/ForestAdmin/forest-rails/pull/785/files#diff-1eae726941328a7d207dbdfd5b67365d6c4f37e62752bdbf68e8d4f1e2a4b4cf) to return `'Date'` for `:timestamptz` and `:timestamp` column types, alongside the existing `:datetime` handling. Adds specs covering all three types.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 49f47f8.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/schema_adapter.rb
+++ b/app/services/forest_liana/schema_adapter.rb
@@ -415,7 +415,7 @@ module ForestLiana
       case column.type
       when :boolean
         type = 'Boolean'
-      when :datetime
+      when :datetime, :timestamptz, :timestamp
         type = 'Date'
       when :date
         type = 'Dateonly'
--- a/spec/services/forest_liana/schema_adapter_spec.rb
+++ b/spec/services/forest_liana/schema_adapter_spec.rb
@@ -1,5 +1,25 @@
 module ForestLiana
   describe SchemaAdapter do
+    describe '#get_type_for' do
+      let(:adapter) { described_class.new(Tree) }
+
+      def column_of_type(type)
+        double('Column', name: 'some_column', type: type, array: false)
+      end
+
+      it 'maps a timestamptz column to the Date type' do
+        expect(adapter.send(:get_type_for, column_of_type(:timestamptz))).to eq 'Date'
+      end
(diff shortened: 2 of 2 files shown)


matthv:
The `:timestamptz` -> `Date` fix looks correct, and the new specs cover the regression well. Two follow-ups worth addressing:

1. `schema_adapter.rb:415` still has no `:timestamp` branch — if `datetime_type` is ever set to `:timestamptz`, plain `timestamp` columns will hit the same silent-drop bug in reverse. Suggest `when :datetime, :timestamptz, :timestamp`.

2. `pie_stat_getter.rb:35-36` only checks `type == :datetime` for groupBy formatting. `timestamptz` columns are now selectable there but will render an unformatted raw key. #787 touches this same block but doesn't fix this either.

Not blocking, but worth a one-line fix here or a fast-follow.

matthv (approved):

PR #786, opened 2026-07-31
fix: skip LIKE scans for malformed UUID searches

ramongtx:
## Problem

When a search term is UUID-shaped but fails the strict `REGEX_UUID` (for example
a truncated or mistyped UUID such as `877a7915-2bc3-4ae-ad4f-5662757713a2`),
`SearchQueryBuilder#search_param` still falls through to
`LOWER(col) LIKE '%…%'` conditions on every text column. On large tables these
force sequential scans across all text columns and can hit the database
statement timeout.

## Fix

Skip the `LIKE` conditions when the search looks UUID-like:

```rb
elsif !(column.respond_to?(:array) && column.array) && text_type?(column.type) && !uuid_like_search?
  conditions << "LOWER(#{column_name}) LIKE :search_value_for_string"
end

# ...

def uuid_like_search?
  @search.match?(/\A[0-9a-f]+-[0-9a-f]+-[0-9a-f]+-[0-9a-f]+-[0-9a-f]+\z/i)
end
```

Exact UUID matches on `id`/`uuid` columns are unaffected (they use the strict
`REGEX_UUID` path).

## Test

Added a spec in `spec/services/forest_liana/search_query_builder_spec.rb`: a
malformed-UUID search produces no `LIKE` condition. Verified the generated SQL
contains `LIKE` before the change and does not after. Full suite green.

### General

### Security

- [x] Consider the security impact of the changes made — reduces load/DoS surface
  from expensive full-table scans; the regex only gates which conditions are built.

> [!NOTE]
> ### Skip LIKE scans in `SearchQueryBuilder` for malformed UUID search strings
> - Adds a `malformed_uuid_search?` predicate in [search_query_builder.rb](https://github.com/ForestAdmin/forest-rails/pull/786/files#diff-fa135f02c0c5311220e1b811b2332b4c7e337954342efc5a94503d847d836841) that returns true when the search string loosely matches a UUID pattern but fails the strict `REGEX_UUID` check.
> - Suppresses `LOWER(col) LIKE '%...%'` conditions on text columns when the search is a malformed UUID, preventing irrelevant full-table LIKE scans across base and associated columns.
> - Returns `@resource.none` when no conditions are built and the search is a malformed UUID, avoiding an unfiltered result set.
> - Behavioral Change: malformed UUID searches now return zero records instead of all records when no other predicates apply.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dde2a07.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -85,7 +85,7 @@ module ForestLiana
             !@resource.defined_enums[column.name][@search.downcase].nil?
             conditions << "#{column_name} =
               #{@resource.defined_enums[column.name][@search.downcase]}"
-          elsif !(column.respond_to?(:array) && column.array) && text_type?(column.type)
+          elsif !(column.respond_to?(:array) && column.array) && text_type?(column.type) && !malformed_uuid_search?
             conditions << "LOWER(#{column_name}) LIKE :search_value_for_string"
           end
         end
@@ -114,7 +114,7 @@ module ForestLiana
               resource = @resource.reflect_on_association(association.to_sym)
               unless (SchemaUtils.polymorphic?(resource))
                 resource.klass.columns.each do |column|
-                  if !(column.respond_to?(:array) && column.array) && text_type?(column.type)
+                  if !(column.respond_to?(:array) && column.array) && text_type?(column.type) && !malformed_uuid_search?
                     if @collection.search_fields.nil? || (association_search &&
                       association_search.include?(column.name))
                       conditions << association_search_condition(resource.table_name,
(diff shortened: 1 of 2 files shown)


matthv:
Found a critical issue: this doesn't skip the LIKE scan, it silently drops the whole filter for the case it targets.

**1. Empty `conditions` returns the unfiltered table.** The final `where` only applies `unless conditions.empty?`. On a table with a `uuid` PK and text columns (the exact case this PR targets), a malformed-UUID search now produces `conditions == []` since no branch matches, so the query runs with no WHERE clause and returns the entire table instead of zero results. That's worse than the original slow query: silently wrong instead of silently slow.

**2. The regex also blocks valid UUID searches on text columns.** It's a superset of `REGEX_UUID`, so a well-formed UUID matches it too. Fine for native `:uuid` columns, but on MySQL/SQLite or UUIDs stored as `varchar`/`text`, valid UUID searches lose matching on those columns.

**3. `searchExtended=1` isn't covered.** Association LIKE scans are still unguarded, so the same timeout risk remains reachable there.

**4. The new test wouldn't catch this.** It only checks `to_sql` excludes `LIKE`, which passes precisely because the WHERE clause disappeared.

Suggest falling back to an explicit `@resource.none` when the guard suppresses LIKE and nothing else matched, tightening the regex, and extending the guard to `searchExtended`.

ramongtx:
Thanks! I've reworked it along your lines. The guard now only trips on malformed UUIDs, falls back to `.none` so nothing-matches returns nothing (not the whole table), and searchExtended is covered too. Specs now assert real row counts.

PR #789, opened 2026-08-11
fix(actions): serve hooks of out-of-mount custom endpoints [PRD-915]

Scra3:
Closes [PRD-915](https://linear.app/forestadmin/issue/PRD-915/trigger-action-step-fails-with-a-generic-data-access-error-for-some).

## Problem

A smart action can declare a custom `endpoint`. Clients — the Forest UI and `agent-client` (workflow executor) — call `<endpoint>/hooks/load` verbatim.

`config/routes/actions.rb` drew those routes **inside the engine**, after `endpoint.sub('forest', '')`. That strip is an absolute→engine-relative conversion: it removes the prefix the mount will add back. It only round-trips when the endpoint starts with `forest`.

An endpoint declared outside the mount (`/domains/organizations/suspensions/suspend`) has nothing to strip, so the mount prepends its own:

```
endpoint   /domains/organizations/suspensions/suspend
route       /forest/domains/organizations/suspensions/suspend/hooks/load
called      /domains/organizations/suspensions/suspend/hooks/load     → 404
```

The route was created without error, at an URL nobody calls. Every form load 404'd, on every client.

It went unnoticed because the only custom-endpoint fixture in the suite (`island.rb`, `forest/custom/islands/enums_action`) starts with `forest/` — the case that round-trips. And until recently `agent-client` swallowed the 404, so the UI degraded to the static form instead of failing.

## Change

- `Model::Action` gains `endpoint_under_forest_mount?`, `engine_relative_endpoint` and `absolute_endpoint`, so the routing decision lives next to where the default endpoint (`forest/actions/…`) is built.
- `config/routes/actions.rb` skips out-of-mount endpoints and uses `engine_relative_endpoint`.
- `ForestLiana::Engine` registers out-of-mount hooks on `Rails.application.routes.append`, at the endpoint verbatim, pointing at `forest_liana/actions#load|change`. Appended from `after_initialize` rather than the routes file: route files are re-evaluated on every reload while appended blocks are never cleared, so registering there would stack a duplicate block per pass. The block body still reads the apimap at each `finalize!`.
- The prefix strip is now anchored. `sub('forest', '')` matched anywhere: `forestry/actions/x` became `/ry/actions/x`, `my/forest/x` became `/my//x`.

PRD-801 semantics are preserved — `load` unconditional, `change` conditional.

`ActionsController` needed no change: it reads only `action_name` (a route default) and `collection_name` from the body, authenticates from the `Authorization` header, and uses no URL helpers or `SCRIPT_NAME`.

## Worth a reviewer's opinion

Out-of-mount hooks routes bypass the engine's `Rack::Cors`, which only wraps the mount. This is already the case for the **execution** endpoint of such actions — the host serves it — so the hook route is consistent with the status quo, and any host declaring an out-of-mount endpoint already needs its own CORS for execution. Flagging it rather than silently accepting it.

## Tests

`458 examples, 0 failures` (was 453).

- `spec/models/forest_liana/model/action_spec.rb` (new) — pins the routing decision and the anchoring, including `forestry/actions/x` and `my/forest/x`.
- `spec/requests/actions_controller_spec.rb` — a new `out_of_mount_action` fixture on `/api/custom/islands/out_of_mount_action`: `/load` and `/change` answer 200 at the URL clients call, and `/forest/api/custom/…/hooks/load` raises `RoutingError`. No change to the dummy app's `routes.rb`, which is what makes it a genuine test of the mechanism.
- Existing in-mount hook specs cover the rewrite of the strip.

Manual check on the dummy app:

```
POST /api/custom/islands/out_of_mount_action/hooks/load   forest_liana/actions#load
POST /custom/islands/enums_action/hooks/load              forest_liana/actions#load   (engine)
```

> [!NOTE]
> ### Fix hook route registration for custom action endpoints outside the Forest engine mount
> - Actions with endpoints outside the `/forest` engine mount previously had no hook routes registered, causing `/hooks/load` and `/hooks/change` to be unserved or caught by host catch-all routes.
> - Adds `endpoint_under_forest_mount?`, `engine_relative_endpoint`, and `absolute_endpoint` helpers to [`ForestLiana::Model::Action`](https://github.com/ForestAdmin/forest-rails/pull/789/files#diff-5f1defc7b847818b67c1a280871d0338f932ae305ac9e00e4095b48b2aeb85e7) to normalize and classify action endpoints.
> - Updates [`config/routes/actions.rb`](https://github.com/ForestAdmin/forest-rails/pull/789/files#diff-2251714258030a8a0e359ad08cbed90a6d11e4f4432cd89054d339ed8b0d6903) to skip drawing engine-relative hook routes for out-of-mount actions.
> - Registers hook routes for out-of-mount actions directly on the host application router via `Rails.application.routes.prepend` in [`lib/forest_liana/engine.rb`](https://github.com/ForestAdmin/forest-rails/pull/789/files#diff-e88f5a727dabeaecc7222321f358a6bc0711d22f75f4e4cc7aa6f467a32cbe1c), ensuring they take precedence over catch-all routes.
>
> 
> #### Changes since #789 opened
>
> - Normalized action endpoints by removing redundant consecutive slashes and stripping leading and trailing slashes in `ForestLiana::Model::Action.normalized_endpoint`, and refactored route registration by extracting inline route-prepending logic for out-of-mount action hooks into a new `ForestLiana::Engine.register_out_of_mount_hook_routes` class method that uses the normalized `absolute_endpoint` values [6658fd2]
> - Added test coverage verifying that redundant and trailing slashes in declared action endpoints are normalized and do not appear in absolute or engine-relative endpoints [6658fd2]
> - Moved endpoint normalization from read-time to write-time in `ForestLiana::Model::Action` by adding a custom `endpoint=` setter that collapses consecutive slashes into a single slash and removes trailing slashes while preserving the leading slash as provided, and simplified the `normalized_endpoint` private method to only strip a single leading slash from the already-normalized stored value. [94bfa2b]
> - Updated `ForestLiana::Model::Action` endpoint routing specs to verify normalization occurs at assignment time and assert on the stored endpoint value. [94bfa2b]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 96359e5.</sup>
> 
>

### DIFF
--- a/app/models/forest_liana/model/action.rb
+++ b/app/models/forest_liana/model/action.rb
@@ -82,6 +82,13 @@ class ForestLiana::Model::Action
     false
   end
 
+  # Normalized at the boundary so every consumer sees one canonical value: the schema sent to
+  # Forest (the UI appends /hooks/load to it), the permission lookup comparing request paths,
+  # and the routes drawn from it. Leading slash kept as declared — consumers tolerate both.
+  def endpoint=(value)
+    @endpoint = value && value.to_s.gsub(%r{/+}, '/').delete_suffix('/')
+  end
+
   def endpoint_under_forest_mount?
     normalized_endpoint.start_with?('forest/')
   end
@@ -97,6 +104,6 @@ class ForestLiana::Model::Action
   private
 
   def normalized_endpoint
-    endpoint.to_s.gsub(%r{/+}, '/').delete_prefix('/').delete_suffix('/')
+    endpoint.to_s.delete_prefix('/')
   end
 end
--- a/app/models/forest_liana/model/action.rb
+++ b/app/models/forest_liana/model/action.rb
@@ -97,6 +97,6 @@ class ForestLiana::Model::Action
   private
 
   def normalized_endpoint
-    endpoint.to_s.delete_prefix('/')
+    endpoint.to_s.gsub(%r{/+}, '/').delete_prefix('/').delete_suffix('/')
   end
 end
--- a/app/models/forest_liana/model/action.rb
+++ b/app/models/forest_liana/model/action.rb
@@ -81,4 +81,22 @@ class ForestLiana::Model::Action
   def persisted?
     false
   end
+
+  def endpoint_under_forest_mount?
+    normalized_endpoint.start_with?('forest/')
+  end
+
+  def engine_relative_endpoint
+    normalized_endpoint.delete_prefix('forest')
+  end
+
+  def absolute_endpoint
+    "/#{normalized_endpoint}"
+  end
+
+  private
+
+  def normalized_endpoint
+    endpoint.to_s.delete_prefix('/')
+  end
 end
--- a/lib/forest_liana/engine.rb
+++ b/lib/forest_liana/engine.rb
@@ -127,6 +127,25 @@ module ForestLiana
           end
         end
       end
+
+      # An endpoint outside the engine mount can only be served by the host router: a route drawn in
+      # the engine gets prefixed with the mount, where no client ever calls. Appended from here and
+      # not from config/routes/actions.rb, which is re-evaluated on every reload while appended
+      # blocks are never cleared — registering there would stack a duplicate block on each pass.
+      Rails.application.routes.append do
+        ForestLiana.apimap.each do |collection|
+          collection.actions.each do |action|
+            next if action.endpoint_under_forest_mount?
+
+            post "#{action.absolute_endpoint}/hooks/load" => 'forest_liana/actions#load',
+              action_name: ActiveSupport::Inflector.parameterize(action.name)
+            if action.hooks && action.hooks[:change].present?
+              post "#{action.absolute_endpoint}/hooks/change" => 'forest_liana/actions#change',
+                action_name: ActiveSupport::Inflector.parameterize(action.name)
+            end
+          end
+        end
+      end
     end
   end
 end
--- a/lib/forest_liana/engine.rb
+++ b/lib/forest_liana/engine.rb
@@ -129,10 +129,11 @@ module ForestLiana
       end
 
       # An endpoint outside the engine mount can only be served by the host router: a route drawn in
-      # the engine gets prefixed with the mount, where no client ever calls. Appended from here and
-      # not from config/routes/actions.rb, which is re-evaluated on every reload while appended
-      # blocks are never cleared — registering there would stack a duplicate block on each pass.
-      Rails.application.routes.append do
+      # the engine gets prefixed with the mount, where no client ever calls. Prepended so a host
+      # catch-all (SPA fallback) cannot shadow the hooks, and registered from here rather than from
+      # config/routes/actions.rb, which is re-evaluated on every reload and would stack a duplicate
+      # block per pass — prepended blocks persist across reloads on their own.
+      Rails.application.routes.prepend do
         ForestLiana.apimap.each do |collection|
           collection.actions.each do |action|
             next if action.endpoint_under_forest_mount?
--- a/lib/forest_liana/engine.rb
+++ b/lib/forest_liana/engine.rb
@@ -128,11 +128,17 @@ module ForestLiana
         end
       end
 
-      # An endpoint outside the engine mount can only be served by the host router: a route drawn in
-      # the engine gets prefixed with the mount, where no client ever calls. Prepended so a host
-      # catch-all (SPA fallback) cannot shadow the hooks, and registered from here rather than from
-      # config/routes/actions.rb, which is re-evaluated on every reload and would stack a duplicate
-      # block per pass — prepended blocks persist across reloads on their own.
+      ForestLiana::Engine.register_out_of_mount_hook_routes
+    end
+
+    # An endpoint outside the engine mount can only be served by the host router: a route drawn in
+    # the engine gets prefixed with the mount, where no client ever calls. Prepended so a host
+    # catch-all (SPA fallback) cannot shadow the hooks, and registered from after_initialize rather
+    # than from config/routes/actions.rb, which is re-evaluated on every reload and would stack a
+    # duplicate block per pass — prepended blocks persist across reloads on their own. Defined here
+    # so the block's closure captures this empty scope, not after_initialize's locals: the block is
+    # retained by the router for the process lifetime and would otherwise pin the bootstrapper.
+    def self.register_out_of_mount_hook_routes
       Rails.application.routes.prepend do
         ForestLiana.apimap.each do |collection|
           collection.actions.each do |action|
--- a/config/routes/actions.rb
+++ b/config/routes/actions.rb
@@ -1,10 +1,13 @@
 ForestLiana.apimap.each do |collection|
   if !collection.actions.empty?
     collection.actions.each do |action|
+      # Endpoints outside the mount are routed on the host app instead — see ForestLiana::Engine.
+      next unless action.endpoint_under_forest_mount?
+
       # Unconditional: clients probe /hooks/load even when no load hook is declared.
-      post action.endpoint.sub('forest', '') + '/hooks/load' => 'actions#load', action_name: ActiveSupport::Inflector.parameterize(action.name)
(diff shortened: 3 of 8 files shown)


Scra3:
Good catch on the append priority — a host SPA catch-all would indeed shadow the hooks and reintroduce the exact 404 this PR fixes. Switched to `prepend` in 74b3bec. Shadowing in the other direction is a non-issue: the only paths taken are `POST <endpoint>/hooks/load|change`, which no host can serve itself (the hooks are lambdas living in the gem). Reload semantics are equivalent — `clear!` re-runs prepended blocks, so they persist across dev reloads just like appended ones.

Scra3:
## Reproduced on unpatched 9.20.9

Ran the reporting client's declaration verbatim — a load hook on `/domains/organizations/suspensions/suspend` — against `origin/main` (9.20.9, the version they run), then against this branch. Same fixture, same request, only the gem changes:

**9.20.9, unpatched**
```
declared endpoint : /domains/organizations/suspensions/suspend
POST /domains/organizations/suspensions/suspend/hooks/load        -> NO ROUTE
POST /forest/domains/organizations/suspensions/suspend/hooks/load -> 200
```

**This branch**
```
POST /domains/organizations/suspensions/suspend/hooks/load        -> 200
POST /forest/domains/organizations/suspensions/suspend/hooks/load -> NO ROUTE
```

The route was always created — just under the mount, where no client calls. Clients (UI and workflow executor alike) call `<endpoint>/hooks/load` verbatim, so every form load 404'd.

The repro fixture isn't in the diff: `out_of_mount_action` already covers the mechanism, and a second fixture with a client-specific endpoint would be redundant.

## On `prepend` vs `append`

Worth noting for whoever reviews the switch: the suite could not originally tell the two apart, since the dummy app had no catch-all. 96359e5 adds a `/api/*path` host fallback so the scenario is actually exercised. Mutation check:

```
prepend -> 32 examples, 0 failures
append  -> 32 examples, 2 failures   (/load and /change)
```

It also surfaced a weak assertion of mine: the `/change` spec only checked for a 200, which the catch-all returns too, so it passed vacuously under `append`. Now asserts the response body.

bexchauveto (commented):
Automated review — 3 findings from a high-effort pass over this branch's diff.

app/models/forest_liana/model/action.rb
```diff
@@ -81,4 +81,22 @@ def initialize(attributes = {})
...
+
+  def endpoint_under_forest_mount?
+    normalized_endpoint.start_with?('forest/')
+  end
+
+  def engine_relative_endpoint
+    normalized_endpoint.delete_prefix('forest')
+  end
+
+  def absolute_endpoint
```
bexchauveto, line 100: **Correctness**: `absolute_endpoint`/`normalized_endpoint` strip only a single leading slash and never a trailing one. An out-of-mount endpoint declared with a trailing slash (`'/domains/.../suspend/'`) or a doubled leading slash (`'//domains/...'`) produces a route path like `.../suspend//hooks/load`, which never matches the client's actual request path — the hook 404s even though the action loads fine in the schema.

lib/forest_liana/engine.rb
```diff
@@ -127,6 +127,26 @@ def eager_load_active_record_descendants app
           end
         end
       end
+
+      # An endpoint outside the engine mount can only be served by the host router: a route drawn in
+      # the engine gets prefixed with the mount, where no client ever calls. Prepended so a host
+      # catch-all (SPA fallback) cannot shadow the hooks, and registered from here rather than from
+      # config/routes/actions.rb, which is re-evaluated on every reload and would stack a duplicate
+      # block per pass — prepended blocks persist across reloads on their own.
+      Rails.application.routes.prepend do
```
bexchauveto, line 142: **Cleanup**: this out-of-mount hooks route registration duplicates `config/routes/actions.rb`'s in-mount version almost verbatim (endpoint expression and controller string aside). A future change to hook routing (new hook type, different `action_name` derivation, added verb) applied to only one of the two blocks would silently make in-mount and out-of-mount actions diverge, with nothing surfacing the drift.

bexchauveto, line 142: **Cleanup**: this `routes.prepend` block is nested inside `config.after_initialize do |app| ... end`, so the Proc's closure captures that block's entire local scope (`app`, `error`, `bootstrapper`) rather than just `ForestLiana.apimap`. The Proc is stored permanently in `Rails.application.routes`' prepend list and re-evaluated on every route reload for the process's lifetime, keeping `bootstrapper` (and the apimap/model state it holds) reachable and un-GC-able long after boot completes. Defining it as a plain method/lambda referencing only `ForestLiana.apimap` would avoid the retention.

app/models/forest_liana/model/action.rb
Scra3: Real, and fixed in 6658fd2 — with one aggravating detail worth recording: the same defect existed **in-mount** before this PR (`engine_relative_endpoint` on a `'forest/x/'` declaration produced `'/x//hooks/load'` inside the engine), so this was not introduced here, but this PR was the right place to kill it for both paths.

`normalized_endpoint` now squeezes repeated slashes and strips the trailing one:

```ruby
endpoint.to_s.gsub(%r{/+}, '/').delete_prefix('/').delete_suffix('/')
```

All three consumers benefit: `absolute_endpoint` ('/domains/x/suspend/' and '//domains/x/suspend' both draw '/domains/x/suspend/hooks/load'), `engine_relative_endpoint` (trailing slash under the mount), and `endpoint_under_forest_mount?` ('//forest/…' is now correctly recognized as in-mount rather than routed on the host).

Spec added covering the four shapes. Full suite green (458 examples).

lib/forest_liana/engine.rb
Scra3: Taken, in 6658fd2. Confirmed real: the block sat inside `config.after_initialize do |app|`, so its closure captured `app`, `error` and `bootstrapper`, and `routes.prepend` retains the block for the process lifetime — the bootstrapper (and the apimap generation state it holds) stayed reachable forever.

Moved to a class method (`ForestLiana::Engine.register_out_of_mount_hook_routes`) called from `after_initialize`. The block's closure now captures that method's empty scope; the only things it references are `ForestLiana.apimap` and constants, resolved at evaluation time. Registration timing is unchanged — same call site, so the apimap is still populated before the router ever evaluates the block.

Scra3: Declining this one, deliberately. The two blocks agree on shape but differ on every load-bearing detail: path base (`engine_relative_endpoint` vs `absolute_endpoint`), controller reference ('actions#load' vs 'forest_liana/actions#load' — engine-relative vs host-absolute), drawing context (engine routes file re-evaluated per reload vs host prepend registered once), and reload semantics — that last difference is precisely what this PR's engine.rb comment documents. A shared helper would need the router, a path lambda and a controller prefix as parameters, which is more machinery than the ~8 duplicated lines and would smear the engine/host boundary that makes each block readable in isolation.

The drift risk is real but covered where it bites rather than at the source: the request specs pin both routings (`spec/requests` exercises in-mount and out-of-mount hooks against the drawn routes), so a change applied to only one block fails a spec instead of silently diverging. If a third hook type ever lands, that is the moment to factor — two call sites is under my threshold, three is not.

app/models/forest_liana/model/action.rb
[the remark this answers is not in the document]
Scra3: Correct, and the blast radius is wider than the finding states — fixed in 94bfa2b.

Beyond a caller appending `/hooks/load`, two in-repo consumers compare against the raw attribute: the permission lookup (`ability/permission.rb:158`) and the schema action finder (`beta_schema_utils.rb:9`) both match the incoming request path against `action.endpoint`. With the route normalized but the attribute raw, a trailing-slash declaration would have routed correctly and then failed permission resolution — a worse failure than the 404, since it appears only after the route starts matching.

Took the first option: the `endpoint=` writer now canonicalizes (squeeze repeated slashes, strip the trailing one, keep the leading slash as declared — both lookup sites tolerate either form), and `normalized_endpoint` reverts to a plain prefix strip. One normalization at the boundary, every consumer — schema, permission lookup, router — sees the same value. Spec asserts the attribute itself is normalized; full suite green (458 examples).

bexchauveto (approved):

PR #790, opened 2026-08-14
fix(has-many-getter): stop dropping records when a has_one is one-to-many in the data

ramongtx:
## Problem

Forest Admin **has-many / related-list** endpoints can silently drop records and return short pages. A page requested with `page[size]=N` comes back with fewer than `N` **distinct** records, so the UI concludes it reached the last page and hides the remaining pages. The total-count badge stays correct, which masks the bug.

## Root cause

`BaseGetter#optimize_record_loading` eager-loads **every** `belongs_to` / `has_one` / `has_and_belongs_to_many` association of the target model into one big `LEFT OUTER JOIN`, and `HasManyGetter#records` then applies `.limit(size).offset(...)` on top of it.

When an association is declared `has_one` but is actually **one-to-many in the data**, Rails keeps it singular and does **not** switch to the limited-ids strategy — so `LIMIT` is applied directly to the row-multiplied JOIN output. A page of N joined rows collapses to only a handful of **distinct** records, and records past that window are dropped. `HasManyGetter#count` stays correct because it counts `DISTINCT` ids.

These associations are eager-loaded only to serialize the related-list columns — not for the query's `WHERE` / `ORDER`. The exceptions are: `SearchQueryBuilder#sort_query` needs the join when sorting BY a relation column (`"rel.field"`), and `FiltersParser` when filtering on a relation (`"rel:field"`, which already self-manages its own join).

## Fix

Override `optimize_record_loading` **in `HasManyGetter`** (the relationship path only — `ResourcesGetter` / index path is untouched) to:

- keep **eager-loaded** only the associations the current request sorts/filters by (parsed from `@params[:sort]` and `@params[:filters]`), so relation-column sort/filter still resolve their joins;
- **preload** the remaining display-only associations — separate queries, no row multiplication.

`LIMIT` then applies to the un-multiplied base rows and returns the correct number of distinct records; display associations are still loaded (via `preload`) for serialization.

The preload stays gated to Rails 7+ (matching #567 — Rails 6 cannot mix `eager_load` and `preload` in the same scope). On Rails 6 the display associations fall back to lazy loading and **pagination is still correct** (the multiplying join is gone either way).

## Tests

New `spec/services/forest_liana/has_many_getter_pagination_spec.rb` adds a target-model `has_one` that resolves to many rows and asserts:

1. a page of size N (< total) returns exactly N **distinct** records, no duplicates;
2. no multiplying JOIN in the paginated SQL for a base-column sort;
3. walking all pages yields every record exactly once;
4. the total count stays correct;
5. sorting BY a relation column still works (relation stays joined);
6. filtering BY a relation column still works (relation stays joined).

Full suite green locally (Ruby 3.3.6 / Rails 6.1): `465 examples, 0 failures`, including the existing `has_many_getter_spec` and `resources_getter_spec`.

## Impact

Affects **every** has-many related list whose target model has a data-wise one-to-many association mis-declared as `has_one` in its eager-load set.

### General

- [x] Write an explicit title for the Pull Request, following Conventional Commits specification

### Security

- [x] Consider the security impact of the changes made — none; this only changes how associations are loaded (eager_load vs preload) for pagination, no change to scoping/authorization or the records returned.

> [!NOTE]
> ### Fix `HasManyGetter` dropping records when `has_one` is one-to-many in data
> - Splits `@includes` into associations that stay eager-loaded (JOINed) versus preloaded, so JOIN row multiplication no longer collapses pagination pages and `LIMIT`/`OFFSET` apply to base rows
> - Keeps an association JOINed only when the request sorts or extended-searches on it; all other display includes move to separate `preload` calls
> - On Rails >= 7, instance-dependent and cross-DB preloads from `analyze_associations` are also preloaded; on older Rails they are not
> - Adds regression tests in [has_many_getter_pagination_spec.rb](https://github.com/ForestAdmin/forest-rails/pull/790/files#diff-5618d656122caa5b96b4a75e95395127f4cd3c601d05c3b90f1d71ee6b724449) covering mis-declared `has_one`, page sizes, distinct IDs, total count, and JOIN behavior for sort and extended search
> - Behavioral Change: `HasManyGetter.optimize_record_loading` now overrides `BaseGetter#optimize_record_loading`; associations no longer referenced in the SQL JOIN may load in separate queries, changing query count and shape
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b69b28.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -121,5 +121,95 @@ module ForestLiana
     def pagination?
       @params[:page] && @params[:page][:number]
     end
+
+    # Overrides BaseGetter#optimize_record_loading for the has-many relationship path.
+    #
+    # BaseGetter eager_loads every belongs_to/has_one/HABTM association of the target
+    # model in a single LEFT OUTER JOIN, and #records then applies LIMIT/OFFSET on top of
+    # it. When an association is declared `has_one` but is one-to-many in the data, Rails
+    # keeps it singular and does NOT switch to the limited-ids strategy, so LIMIT is applied
+    # directly to the row-multiplied JOIN output: a page collapses to fewer DISTINCT records
+    # than requested and records past the window are silently dropped (the count stays right
+    # because #count uses COUNT(DISTINCT id)).
+    #
+    # These associations are eager-loaded only to serialize the related-list columns, not
+    # for the query's WHERE/ORDER. So keep eager-loaded only the associations the current
+    # request sorts or filters BY (their joins must resolve), and preload the remaining
+    # display-only associations (separate queries, no row multiplication). LIMIT then applies
+    # to the un-multiplied base rows and returns the correct DISTINCT records.
+    def optimize_record_loading(resource, records, force_preload = true)
+      polymorphic, preload_loads = analyze_associations(resource)
+      display_includes = @includes.uniq - preload_loads - polymorphic - @optional_includes
+
+      keep_eager = display_includes & associations_referenced_by_sort_and_filters
+      move_to_preload = display_includes - keep_eager
+
+      result = records.eager_load(keep_eager)
+
+      # NOTICE: Rails 6 cannot mix `eager_load` and `preload` in the same scope (see #567),
+      # so the display-only associations are preloaded on Rails 7+ and lazy-loaded on Rails 6.
+      # Either way LIMIT is no longer applied to a row-multiplied JOIN.
+      if Rails::VERSION::MAJOR >= 7 && force_preload
+        result = result.preload(move_to_preload + preload_loads)
+      end
+
+      result
+    end
+
+    # Association names (symbols) referenced by the request's sort and filters. Their joins
+    # must stay eager-loaded so SearchQueryBuilder#sort_query (ORDER BY a relation column)
+    # and relation filters resolve their columns.
+    def associations_referenced_by_sort_and_filters
+      (associations_referenced_by_sort + associations_referenced_by_filters).uniq
+    end
+
+    # @params[:sort] is a comma-separated string; a leading '-' means descending and a '.'
+    # references a relation (the part before the dot). e.g. '-owner.name' => :owner
+    def associations_referenced_by_sort
+      sort = @params[:sort]
+      return [] if sort.nil? || sort.to_s.empty?
+
+      sort.to_s.split(',').map do |field|
+        field = field.strip.sub(/\A-/, '')
+        field.include?('.') ? field.split('.').first.to_sym : nil
+      end.compact
+    end
+
+    # @params[:filters] is a JSON string or a condition tree (Hash). A leaf field references
+    # a relation with ':' (belongs_to style) or '.' (nested); aggregation nodes nest under
+    # 'conditions'. e.g. 'owner:name' => :owner
+    def associations_referenced_by_filters
+      filters = @params[:filters]
+      return [] if filters.nil? || (filters.respond_to?(:empty?) && filters.empty?)
+
+      tree = filters.is_a?(String) ? parse_filters_json(filters) : filters
+      tree = tree.to_unsafe_h if tree.respond_to?(:to_unsafe_h)
+      extract_filter_associations(tree)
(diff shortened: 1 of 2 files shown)


bexchauveto:
## Review

The core idea is sound and the pagination bug is real, but the diff drops joins another feature depends on.

### 1. `app/services/forest_liana/has_many_getter.rb:147` — extended search on a related list now 500s

`SearchQueryBuilder#search_param` emits `... OR LOWER("users"."name") LIKE ...` for every association in `@includes` when `searchExtended=1`, and it relied on `optimize_record_loading`'s `eager_load` to supply the `LEFT OUTER JOIN`. Those associations are referenced by neither `@params[:sort]` nor `@params[:filters]`, so they now land in `move_to_preload` — and `preload` never joins, on any Rails version.

Reproduced (Island → trees, `search=alice`, `searchExtended=1`):
- on `main`: passes, SQL contains the joins;
- on this branch: `SQLite3::SQLException: no such column: users.name` (Postgres: *missing FROM-clause entry for table "users"*). Generated SQL: `... OR LOWER("users"."name") LIKE '%alice%' ... FROM "trees"`, no JOIN at all.

This breaks the related-list search and its `/count`, and also `ResourcesGetter.get_ids_from_request` (“select all records” on a related list whose subset query carries an extended search). The same hazard applies to customer smart-field `search:` lambdas that reference an included association's table.

Suggested fix: when `@params[:search]` is present and `@params['searchExtended'].to_i == 1`, keep `display_includes` eager (or intersect with the associations `SearchQueryBuilder` actually produced conditions for).

### 2. `app/services/forest_liana/has_many_getter.rb:152` — on Rails 6, display associations are neither eager-loaded nor preloaded (N+1)

The `Rails::VERSION::MAJOR >= 7` gate is copied from #567, but the Rails 6 limitation it guards is *preloading instance-dependent scopes*, not mixing `eager_load` with `preload`. Verified on this repo's Rails 6.1.7.9:

- `Tree.eager_load(:owner).preload(:island).to_a` works and marks both associations loaded;
- `Island.preload(:eponymous_tree)` raises `ArgumentError: The association scope 'eponymous_tree' is instance dependent ... Preloading instance dependent scopes is not supported`.

`move_to_preload` contains only plain associations (instance-dependent and cross-DB ones were already subtracted into `preload_loads`), so it can be preloaded unconditionally; only `preload_loads` needs the version gate. As written, a Rails 6 related list with `page[size]=100` and 5 `belongs_to` display columns goes from 1 query to ~501 during serialization.

### 3. `app/services/forest_liana/has_many_getter.rb:181` — `associations_referenced_by_filters` is dead weight (~45 lines)

`FiltersParser#apply_filters` already does `@resource = @resource.eager_load(join.name)` for every `field` containing `:`, and it runs inside `@search_query_builder.perform(association)` — i.e. *before* `optimize_record_loading` receives the relation — so the join is already on the scope. The `.`-notation branch never fires either: `FiltersParser#is_belongs_to` only recognizes `:`, and a dotted field is resolved as a base-table column (raising `Field '...' not found`). Only `associations_referenced_by_sort` is load-bearing here; dropping the filter helpers also removes a second, divergent copy of `ResourcesGetter#extract_associations_from_filter`.

### Verified as OK

- Sorting by a relation column still resolves (`keep_eager` covers it), and aliasing actually improves: with only the sorted association joined, `SearchQueryBuilder#detect_reference`'s real-table-name assumption stops silently sorting by the wrong join (e.g. `cutter.name` previously emitted `users.name`, which was the `owner` join).
- `count`, `query_for_batch`, polymorphic / cross-DB / instance-dependent handling and the `force_preload` argument are unchanged in behaviour; `HasManyGetter` is the only caller of the override.
- The spec cleanup of the dynamically added `has_one :island_neighbour` does fully revert. Full suite shows an identical failure set on `main` (458 examples, 40 failures) and on this branch (465 examples, 40 failures) — those 40 are pre-existing/environmental — and the 7 new examples pass.
- Nit: `Tree.undef_method(...)` in `after(:each)` is a private-method call swallowed by `rescue nil`, so it is a no-op; the reflection deletes are what actually clean up.

PR #791, opened 2026-08-24
fix(count): search the same columns for a count as for its list

hercemer42:
Closes [PRD-1034](https://linear.app/forestadmin/issue/PRD-1034/forest-rails-extended-search-returns-no-count-500-and-an-inflated)

## What breaks today

With extended search on, against any `forest_liana` project:

- **Top-level list — the count is gone.** `GET /forest/:collection/count` returns **500** (`PG::UndefinedTable: missing FROM-clause entry`). The table lists the right records; only the count fails.
- **Related data — the count is too high, silently.** The relationship tab lists N records and announces more than N.

Nothing changed in the gem. The front stopped sending `fields[...]` on `/count` in 2.1245.0 (13 Aug), correctly reasoning that a count never serializes records. `forest_liana` derives **search behaviour** from that parameter, not just serialization: `fields` is what armed the JOINs the extended-search predicates need. The hole dates to the 2017 count optimisation (#117) and sat latent for eight years because the front always sent `fields`.

Note this also restores a documented recipe: `/legacy/ruby-agent/.../override-a-route` tells you to set `params['searchExtended'] = '1'` on **both** `index` and `count`. That recipe 500s on `main`.

## The fix

Root cause of everything below: a single `@includes` was serving three different jobs — which tables the search may name, which the query eager-loads, and which get serialized. They are not the same set.

**`resources_getter.rb` — the 500.** `@count_needs_includes` was set as a side effect of `extract_associations_from_filter`, which `searchExtended=1` never reaches (`compute_includes` takes the `else` branch, and `field_names_requested` early-returns on the missing `fields`). It now comes from the params in `initialize`, so `count` eager-loads and the association columns in the `WHERE` have a `FROM` entry.

`!@params[:search].nil?` rather than `.present?` is deliberate — `search=''` still builds `LIKE '%%'` predicates (`search_query_builder.rb:60` guards on `if @search`, and `''` is truthy), so `present?` would leave that case broken. A test pins it.

**`has_many_getter.rb` — the inflated count.** `compute_includes` narrowed the association set by the projection only when that set was non-empty, so a count with no projection searched every association while the list searched only the projected ones. Under `searchExtended=1` the footprint is now projection-independent, mirroring `ResourcesGetter#compute_includes`.

**`has_many_getter.rb` — the two sets that must not be shared.** Making the footprint projection-independent surfaced two things the first commit got wrong, both caught in review and fixed in 575f05a:

- `SearchQueryBuilder` now receives `searchable_includes` — `@includes` minus what `analyze_associations` rejects (polymorphic, cross-database, instance-dependent scopes). Without this, a related list with a cross-database association returned **500**: the predicate named a table `optimize_record_loading` had dropped from the eager load. `@includes` itself is untouched, so those associations keep their `preload`.
- `includes_for_serialization` is overridden to clamp back to the projection, as `ResourcesGetter` already did. Without it the related list answered with every association plus a populated `included` array, ignoring the `fields[...]` the client sent.

## Behaviour change worth a look

**The related list returns 2 where it returned 1** for an extended search carrying a projection. The count was always right; the list was under-searching. Making the footprint projection-independent costs the related list an eager-load of all one-associations under extended search — the cost the top-level list already pays. Narrowing the count instead isn't available: the count has no projection left to narrow by.

Outside the extended-search path nothing moves. `@count_needs_includes` can now only go false→true, never the reverse, so no existing count path loses its eager-load. And `Calculations#calculate` forces `COUNT(DISTINCT pk)` on any eager-loaded relation, so the widened set cannot inflate a count — including via the HABTM member of `SUPPORTED_ASSOCIATION_MACROS`, which the dummy app has no model to exercise.

## Tests

There was **no coverage of `searchExtended=1` anywhere in the repo** — every spec passed `'0'`. That is why this shipped. 28 tests added, written red before each fix.

| | revert `resources_getter` | revert `has_many_getter` (a11cc22) | revert `has_many_getter` (575f05a) |
| -- | -- | -- | -- |
| `count_spec.rb` | 4 fail | 2 fail | — |
| `resources_getter_spec.rb` | 1 fail | 0 | — |
| `has_many_getter_spec.rb` | 0 | 3 fail | 3 fail |

Each fix is independently guarded. Where a scenario has both a count and a list, the test asserts `count == list length` **and** the absolute number, so a wrong-but-consistent pair can't pass.

Verified unaffected, not assumed: `composite_primary_keys`, `resources_getter_composite_keys`, `filters_parser`, `search_query_builder`.

## Verified on PostgreSQL

The whole suite runs on sqlite, and the reported symptom is a PG error — so both symptoms were reproduced on a local Postgres 15 rig and confirmed fixed, merge-base vs branch, same rig and data:

| Request (`search=kiwi`, `searchExtended=1`) | `c416ff7` | `575f05a` |
| -- | -- | -- |
| `GET /forest/Tree/count` — no `fields` | **500** | `{"count":2}` |
| `GET /forest/Island/1/relationships/trees/count` — no `fields` | `{"count":2}` | `{"count":2}` |
| `GET /forest/Island/1/relationships/trees?fields[Tree]=id,name` | **1 row** | **2 rows**, keys `[id, name]` |

Also checked there: `count` + `sort=-id` returns 200 — PostgreSQL rejects `SELECT DISTINCT … ORDER BY <expr outside the select list>`, and this PR routes counts through `optimized_count` in cases that previously skipped it. SQLite cannot expose that failure class.

And a genuine cross-engine case (PG primary + SQLite secondary): the related list and count both return 200, while a search matching only via `manufacturer.name` still returns `{"count":1}` — the search set was narrowed to joinable tables without narrowing too far.

## Not in this PR

- `has_many` association filters (`trees_owned:name`) 500 with `ambiguous column name` on both the list and the count, regardless of `searchExtended`. Different defect, seen only against the dummy models.
- A collection declaring `search_fields` with a dotted `has_many` entry still 500s on `/count` under extended search — pre-existing, static analysis only.
- `query_for_batch` has the same missing-join hole one accessor over, reachable from a bulk action over "select all records" — pre-existing.
- An ADR is owed for choosing to widen the list's footprint over implementing the `Forest-Projection` header (PRD-928). `docs/adr/**` lives in another repo.

> [!NOTE]
> ### Fix count queries to search the same columns as list queries
> - Adds `BaseGetter#searchable_includes` to compute only the associations that eager loading actually joins (excluding polymorphic and preload-only associations).
> - `HasManyGetter` and `ResourcesGetter` now pass `searchable_includes` to `SearchQueryBuilder` instead of the raw includes list, so counts search the same columns as their corresponding listings.
> - When `searchExtended` is enabled (`searchExtended=1`), includes are no longer narrowed by the requested field projection; the full set of one-to-one associations is retained for both count and list.
> - `ResourcesGetter` constructor now sets `@count_needs_includes = true` whenever a search parameter is present, replacing a redundant assignment in `extract_associations_from_filter`.
> - Behavioral Change: count queries for searches now include association joins they previously lacked, which may change count results for filtered searches; reviewers should check `ResourcesGetter#initialize` and `HasManyGetter#initialize`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9e553f8.</sup>
> 
>

### DIFF
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -14,7 +14,7 @@ module ForestLiana
       @field_names_requested = field_names_requested
       @collection = get_collection(@collection_name)
       compute_includes()
-      @search_query_builder = SearchQueryBuilder.new(@params, searchable_includes, @collection, forest_user)
+      @search_query_builder = SearchQueryBuilder.new(@params, searchable_includes(model_association), @collection, forest_user)
 
       prepare_query()
     end
@@ -64,14 +64,6 @@ module ForestLiana
     end
 
     private
-
-    # A search predicate can only name an association the eager load actually joins;
-    # analyze_associations drops the rest, leaving their table out of the FROM clause.
-    def searchable_includes
-      polymorphic, preload_loads = analyze_associations(model_association)
-
-      (@includes - polymorphic - preload_loads).map(&:to_sym)
-    end
 
     def compute_includes
       @optional_includes = []
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -76,12 +76,18 @@ module ForestLiana
             inclusion = SchemaUtils.model_included?(association.klass)
           end
 
-          if @field_names_requested.any?
+          if @field_names_requested.any? && !search_extended?
             inclusion && @field_names_requested.include?(association.name)
           else
             inclusion
           end
         end.map(&:name)
+    end
+
+    # Extended search reaches the associated tables, so its footprint must not depend
+    # on the projection: a count carries none, and would otherwise search more than the list.
+    def search_extended?
+      @params['searchExtended'].to_i == 1
     end
 
     def field_names_requested
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -14,8 +14,7 @@ module ForestLiana
       @field_names_requested = field_names_requested
       @collection = get_collection(@collection_name)
       compute_includes()
-      includes_symbols = @includes.map { |include| include.to_sym }
-      @search_query_builder = SearchQueryBuilder.new(@params, includes_symbols, @collection, forest_user)
+      @search_query_builder = SearchQueryBuilder.new(@params, searchable_includes, @collection, forest_user)
 
       prepare_query()
     end
@@ -58,7 +57,21 @@ module ForestLiana
       @records.limit(limit).offset(offset)
     end
 
+    def includes_for_serialization
+      return super if @field_names_requested.empty?
+
+      super & @field_names_requested.map(&:to_s)
+    end
+
     private
+
+    # A search predicate can only name an association the eager load actually joins;
+    # analyze_associations drops the rest, leaving their table out of the FROM clause.
+    def searchable_includes
+      polymorphic, preload_loads = analyze_associations(model_association)
+
+      (@includes - polymorphic - preload_loads).map(&:to_sym)
+    end
 
     def compute_includes
       @optional_includes = []
--- a/app/services/forest_liana/base_getter.rb
+++ b/app/services/forest_liana/base_getter.rb
@@ -24,6 +24,14 @@ module ForestLiana
       @optional_includes = []
     end
 
+    # A search predicate can only name an association the eager load actually joins;
+    # analyze_associations drops the rest, leaving their table out of the FROM clause.
+    def searchable_includes(resource)
+      polymorphic, preload_loads = analyze_associations(resource)
+
+      (@includes - polymorphic - preload_loads).map(&:to_sym)
+    end
+
     def optimize_record_loading(resource, records, force_preload = true)
       polymorphic, preload_loads = analyze_associations(resource)
       result = records.eager_load(@includes.uniq - preload_loads - polymorphic - @optional_includes)
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -6,7 +6,7 @@ module ForestLiana
       @resource = resource
       @params = params
       @user = forest_user
-      @count_needs_includes = false
+      @count_needs_includes = !@params[:search].nil?
       @collection_name = ForestLiana.name_for(@resource)
       @collection = get_collection(@collection_name)
       @fields_to_serialize = get_fields_to_serialize
@@ -176,8 +176,6 @@ module ForestLiana
           @count_needs_includes = true
         end
       end
-
-      @count_needs_includes = true if @params[:search]
 
       associations.uniq
     end
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -13,7 +13,7 @@ module ForestLiana
       @field_names_requested = field_names_requested
       @segment = get_segment
       compute_includes
-      @search_query_builder = SearchQueryBuilder.new(@params, @includes, @collection, @user)
+      @search_query_builder = SearchQueryBuilder.new(@params, searchable_includes(@resource), @collection, @user)
 
       prepare_query
       @base_records_for_batch = @records
--- a/spec/requests/count_spec.rb
+++ b/spec/requests/count_spec.rb
@@ -130,3 +130,246 @@ describe 'Requesting Owner', :type => :request  do
     end
   end
 end
+
+describe 'Requesting Tree count with extended search', :type => :request do
+  let(:scope_filters) { { 'scopes' => {}, 'team' => { 'id' => '1', 'name' => 'Operations' } } }
+
+  # NOTICE: `Kiwi Tree` matches the search on its own column, `Lemon Tree` matches
+  #         only through `owner.name`, and `Apple Tree` matches nothing. Extended
+  #         search must therefore return 2 records, plain search 1.
+  before(:each) do
+    Tree.destroy_all
+    Location.destroy_all
+    Island.destroy_all
+    User.destroy_all
+
+    michel = User.create(name: 'Michel')
+    kiwi_grower = User.create(name: 'Kiwi Grower')
+    @island = Island.create(name: 'Home Island')
+
+    Tree.create(name: 'Kiwi Tree', owner: michel, cutter: michel, island: @island)
+    Tree.create(name: 'Lemon Tree', owner: kiwi_grower, cutter: michel, island: @island)
+    Tree.create(name: 'Apple Tree', owner: michel, cutter: michel, island: @island)
+
+    allow(ForestLiana::IpWhitelist).to receive(:retrieve) { true }
+    allow(ForestLiana::IpWhitelist).to receive(:is_ip_whitelist_retrieved) { true }
+    allow(ForestLiana::IpWhitelist).to receive(:is_ip_valid) { true }
+
+    allow_any_instance_of(ForestLiana::Ability).to receive(:forest_authorize!) { true }
+
+    allow(ForestLiana::ScopeManager).to receive(:fetch_scopes).and_return(scope_filters)
+  end
+
+  after(:each) do
+    Tree.destroy_all
+    Location.destroy_all
+    Island.destroy_all
+    User.destroy_all
+  end
+
+  token = JWT.encode({
+                       id: 38,
+                       email: 'michael.kelso@that70.show',
+                       first_name: 'Michael',
+                       last_name: 'Kelso',
+                       team: 'Operations',
+                       rendering_id: 16,
+                       exp: Time.now.to_i + 2.weeks.to_i,
+                       permission_level: 'admin'
+                     }, ForestLiana.auth_secret, 'HS256')
+
+  headers = {
+    'Accept' => 'application/json',
+    'Content-Type' => 'application/json',
+    'Authorization' => "Bearer #{token}"
+  }
+
+  def count_returned
+    JSON.parse(response.body)['count']
+  end
+
+  def records_returned
+    JSON.parse(response.body)['data'].size
+  end
+
+  describe 'count' do
+    let(:projection) { { 'Tree' => 'id,name,owner' } }
+    let(:params) {
+      {
+        page: { 'number' => '1', 'size' => '10' },
+        search: 'kiwi',
+        searchExtended: '1',
+        timezone: 'Europe/Paris'
+      }
+    }
+
+    describe 'when the request carries no fields' do
+      it 'responds 200 with the number of records the list returns for the same search' do
+        get '/forest/Tree', params: params, headers: headers
+        listed = records_returned
+
+        get '/forest/Tree/count', params: params, headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(2)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when the request carries fields' do
+      it 'responds 200 with the number of records the list returns for the same search' do
+        get '/forest/Tree', params: params.merge(fields: projection), headers: headers
+        listed = records_returned
+
+        get '/forest/Tree/count', params: params.merge(fields: projection), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(2)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when the same count is requested twice' do
+      it 'returns the same count both times' do
+        get '/forest/Tree/count', params: params, headers: headers
+        first_count = count_returned
+
+        get '/forest/Tree/count', params: params, headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(2)
+        expect(count_returned).to eq(first_count)
+      end
+    end
+
+    describe 'when extended search is off' do
+      it 'counts only the records matching on their own columns' do
+        get '/forest/Tree', params: params.merge(searchExtended: '0'), headers: headers
+        listed = records_returned
+
+        get '/forest/Tree/count', params: params.merge(searchExtended: '0'), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(1)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when searchExtended is absent from the request' do
+      it 'counts only the records matching on their own columns' do
+        get '/forest/Tree/count', params: params.except(:searchExtended), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(1)
+      end
+    end
+
+    describe 'when the search matches no record at all' do
+      it 'responds 200 with a count of zero' do
+        get '/forest/Tree', params: params.merge(search: 'sequoia'), headers: headers
+        listed = records_returned
+
+        get '/forest/Tree/count', params: params.merge(search: 'sequoia'), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(0)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when the search is an empty string' do
+      it 'counts every record' do
+        get '/forest/Tree/count', params: params.merge(search: ''), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(3)
+      end
+    end
+
+    describe 'when the request carries no search at all' do
+      it 'counts every record' do
+        get '/forest/Tree/count', params: params.except(:search), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(3)
+      end
+    end
+  end
+
+  describe 'count on relationships' do
+    let(:projection) { { 'Tree' => 'id,name' } }
+    let(:params) {
+      {
+        page: { 'number' => '1', 'size' => '10' },
+        search: 'kiwi',
+        searchExtended: '1',
+        timezone: 'Europe/Paris'
+      }
+    }
+
+    describe 'when the count carries no fields and the list carries the projection' do
+      it 'counts the number of records the related list returns' do
+        get "/forest/Island/#{@island.id}/relationships/trees",
+            params: params.merge(fields: projection), headers: headers
+        listed = records_returned
+
+        get "/forest/Island/#{@island.id}/relationships/trees/count", params: params, headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(2)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when both the count and the list carry the projection' do
+      it 'counts the number of records the related list returns' do
+        get "/forest/Island/#{@island.id}/relationships/trees",
+            params: params.merge(fields: projection), headers: headers
+        listed = records_returned
+
+        get "/forest/Island/#{@island.id}/relationships/trees/count",
+            params: params.merge(fields: projection), headers: headers
+
+        expect(response.status).to eq(200)
+        expect(count_returned).to eq(2)
+        expect(count_returned).to eq(listed)
+      end
+    end
+
+    describe 'when extended search is off' do
+      it 'counts only the related records matching on their own columns' do
+        get "/forest/Island/#{@island.id}/relationships/trees",
+            params: params.merge(searchExtended: '0', fields: projection), headers: headers
+        listed = records_returned
+
+        get "/forest/Island/#{@island.id}/relationships/trees/count",
+            params: params.merge(searchExtended: '0'), headers: headers
+
+        expect(response.status).to eq(200)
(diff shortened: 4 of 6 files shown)


app/services/forest_liana/has_many_getter.rb
```diff
@@ -76,14 +76,20 @@ def compute_includes
             inclusion = SchemaUtils.model_included?(association.klass)
           end
 
-          if @field_names_requested.any?
+          if @field_names_requested.any? && !search_extended?
```
hercemer42, line 84: Claude Opus 5 (claude-opus-5[1m]): `Must fix`

A related-data list that works today returns **500** after this change, whenever the related collection has a one-association that `eager_load` refuses — a cross-database one, or one carrying an instance-dependent scope (`->(record) { … }`).

Reproduced on this branch with the repo's own dummy models:

`GET /forest/Manufacturer/:id/relationships/products?search=zzz&searchExtended=1&fields[Product]=id,name`
— **500** at `a11cc22`, **200** with this file reverted to the merge base. (`Product belongs_to :driver`, and `Driver < UserRecord` connects to a separate database.)

Mechanism: `compute_includes` now hands the un-narrowed set to `SearchQueryBuilder` (`has_many_getter.rb:17-18`), which emits `LOWER(drivers.firstname) LIKE …` for every association in `@includes` (`search_query_builder.rb:113`). But `BaseGetter#optimize_record_loading` subtracts `preload_loads` before eager-loading (`base_getter.rb:29`), and `analyze_associations` puts exactly those two kinds of association there (`base_getter.rb:36-46`). The predicate lands on a table that never gets joined.

`ResourcesGetter#compute_includes:124-127` guards half of this with `.reject { |association| separate_database?(...) }`; this method guards neither. The dummy app masks it for `Tree` only by luck — `eponymous_island` is instance-dependent and dropped from the eager-load, but resolves against the `isle` join that `:island` already contributes.

Rejecting the same associations `analyze_associations` will reject, before `@includes` reaches `SearchQueryBuilder`, would close it.

hercemer42, line 84: Claude Opus 5 (claude-opus-5[1m]): `Must fix`

**Applies to:** the serialized response — `app/services/forest_liana/base_getter.rb:13-18` and `app/controllers/forest_liana/associations_controller.rb:133-137`, neither in this diff. The widened `@includes` on this line is the cause.

The related list now ignores the projection the client asked for. Reproduced on this branch:

`GET /forest/Island/:id/relationships/trees?search=kiwi&searchExtended=1&fields[Tree]=id,name`
→ rows come back carrying `relationships: ["owner","cutter","island","eponymous_island","location"]` and a populated `included` array, where the client asked for two attributes.

`HasManyGetter` does not override `includes_for_serialization`, so it inherits `BaseGetter`'s, which returns `@includes` verbatim. `ResourcesGetter` *does* override it (`resources_getter.rb:131-133`) and clamps back with `super & @fields_to_serialize`. The controller then appends the whole set to `fields_to_serialize` and passes it as `include:`.

Widening the *search* footprint is the intent of this PR; widening the *response payload* is a side effect of sharing one `@includes` between the two concerns. On a collection with a HABTM — also in `SUPPORTED_ASSOCIATION_MACROS` — it serialises an entire to-many relationship per row.

None of the new specs would catch it: `count_spec.rb` asserts only `data.size`, never the serialized field set.

hercemer42 (commented):
**Spec (PRD-1034):** both acceptance criteria are met — the count equals the list's row count for the same search, and the extended-search footprint no longer depends on the projection. `resources_getter.rb` is clean: `!@params[:search].nil?` is the right predicate (`.present?` would reintroduce the 500 for `search=''`, since `search_query_builder.rb:60` guards on `if @search` and `''` is truthy), the flag is monotonically ≥ its old value so no count path loses its eager-load, and `Calculations#calculate` forces `COUNT(DISTINCT pk)` on any eager-loaded relation so the widened set cannot inflate a count.

`has_many_getter.rb` reaches the same criterion in a way that breaks two things the ticket did not scope. Both are inline, both reproduce on this branch with the repo's own dummy models, and both trace to sharing one `@includes` between the search footprint and the eager-load/serialization sets.

---

Claude Opus 5 (claude-opus-5[1m]): `Should fix`

**Applies to:** the PR as a whole

Whoever implements the `Forest-Projection` header (PRD-928) will have to decide whether the related list goes back to honouring the projection, and nothing in the tree will tell them why it stopped. This PR chooses to widen the list's search footprint rather than narrow the count — because narrowing needs the header forest-rails does not implement — and that costs existing clients extra JOINs and a changed row count. Hard to reverse once shipped, surprising without the front-release context, and a genuine trade-off. An org-wide ADR search (6 records, control query passing) found nothing covering count-route search-footprint semantics, and this PR adds none. Consider `/adr`.



---

Claude Opus 5 (claude-opus-5[1m]): `Preferential`

**Applies to:** `app/services/forest_liana/resources_getter.rb:100-127` (not in this diff)

A collection declaring `search_fields ['name', 'cuttings.label']` — `cuttings` being a `has_many` — still returns 500 on `/count` under `searchExtended=1`, so "extended search no longer 500s on count" is narrower than it reads. `search_query_builder.rb:129-150` emits predicates against many-association tables named in `search_fields`, and that block sits inside the `searchExtended == 1` gate; but `includes_for_smart_search` (`resources_getter.rb:105-113`), the only thing that would join them, is computed solely in the `searchExtended != 1` branch. Pre-existing, not a regression from this PR. Static analysis only — the dummy app declares no `search_fields`, so I could not execute it.



---

Claude Opus 5 (claude-opus-5[1m]): `Preferential`

**Applies to:** `app/services/forest_liana/resources_getter.rb:19` and `:54-55` (not in this diff)

A bulk smart action using "select all records" with extended search on hits the same missing-join 500 this PR fixes for `count`, one accessor over. `@base_records_for_batch = @records` is captured straight out of `prepare_query`, before anything eager-loads, and `get_ids_from_request` → `fetch_ids` → `query_for_batch.find_in_batches` calls neither `perform` nor `optimized_count`. Pre-existing and outside this diff; flagging it because it is the same defect, and the front does put `searchExtended` in `allRecordsSubsetQuery`. Static analysis only.

app/services/forest_liana/has_many_getter.rb
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Agreed, and fixed in 575f05a.

The search set and the eager-load set are now the same set. `SearchQueryBuilder` gets `searchable_includes` — `@includes` minus what `analyze_associations` rejects (polymorphic, cross-database, instance-dependent) — so a predicate can only name a table the `eager_load` actually joins. `@includes` itself is untouched, so cross-database associations keep their `preload` and the serialization set is unaffected.

Verified on this branch with the repo's own dummy models, both routes, not just the one in the finding:

| `search=zzz&searchExtended=1&fields[Product]=id,name` | a11cc22 | 575f05a |
| -- | -- | -- |
| `GET /forest/Manufacturer/:id/relationships/products` | 500 | 200 |
| `GET /forest/Manufacturer/:id/relationships/products/count` | 500 | 200 |

The count route was 500ing for this shape before my first commit too, so this closes a case the original fix left open rather than only undoing a regression.

Two specs pin it, both red at a11cc22: one asserting no raise when the search cannot reach the cross-database column, one asserting the search still matches through `manufacturer` — the second is there so a future over-narrowing of the set fails loudly instead of silently returning 0.

Not changed: `ResourcesGetter#compute_includes:124-127` rejects `separate_database?` but not instance-dependent associations, so it carries the same latent hole. Pre-existing, not reachable in the dummy app (`eponymous_island` resolves against the `isle` join `:island` contributes), and out of scope here — flagging it rather than folding it in.

hercemer42: Claude Opus 5 (claude-opus-5[1m]): Agreed, and fixed in 575f05a.

`HasManyGetter` now overrides `includes_for_serialization` and clamps to the projection, the way `ResourcesGetter` already did:

```ruby
def includes_for_serialization
  return super if @field_names_requested.empty?

  super & @field_names_requested.map(&:to_s)
end
```

The early return matters: without it, a related list carrying no projection would start serializing nothing, where today it serializes everything. That is a live case now that the front sends projections via the `Forest-Projection` header, which this gem does not read — so a no-`fields` related list is what it actually receives. The fix restores the pre-PR behaviour on both sides rather than only the one the finding named.

Verified at request level on this branch — `GET /forest/Island/:id/relationships/trees?search=kiwi&searchExtended=1&fields[Tree]=id,name`:

| | a11cc22 | 575f05a |
| -- | -- | -- |
| `attributes` | `["id","name"]` | `["id","name"]` |
| `relationships` | `["owner","cutter","island","eponymous_island","location"]` | `nil` |
| `included` | 2 records | 0 |

Two specs pin it, on `includes_for_serialization` directly — that is the exact method `associations_controller.rb:133` consumes — one for the clamped case (red at a11cc22) and one guarding the no-projection case.

You were right that nothing would have caught this: the assertions I added in the first commit only ever looked at `data.size`.

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): Responses to the three findings in the review body (they have no line anchor, so they get a comment rather than a thread).

Both `Must fix` findings are fixed in 575f05a and answered in their threads. The three below are not code changes to this PR, and I am not making them here — reasons per item.

---

**`Should fix` — an ADR is owed. Agreed, but it cannot land in this PR.**

The decision is real and still stands after 575f05a: under `searchExtended=1` the search footprint is projection-independent and the related list eager-loads every one-association, chosen over implementing the `Forest-Projection` header. Fixing the serialization clamp narrowed the blast radius — the response payload is back to what the client asked for — but the JOIN cost and the count/list reconciliation strategy are unchanged, which is what the ADR would record.

It cannot be a commit here: `docs/adr/**` lives in `ForestAdmin/forestadmin`, not in this repo. It needs `/adr` run there, which is a separate change with a separate reviewer. Flagging for the author to decide rather than silently dropping it.

---

**`Preferential` — dotted `has_many` `search_fields` still 500 on count. Correct, and deliberately not fixed here.**

`search_query_builder.rb:129-150` emits predicates against many-association tables named in `search_fields`, inside the `searchExtended == 1` gate; `includes_for_smart_search` (`resources_getter.rb:105-113`), the only thing that would join them, is computed solely in the `searchExtended != 1` branch. Pre-existing, unchanged by either commit, and a different mechanism from the one this ticket describes.

Worth noting it is already known: the ticket's own workaround section tells affected customers to declare `search_fields` with own-column entries only, no `assoc.field` entries. That workaround exists precisely because this shape is broken. It deserves its own ticket, not a fold-in — fixing it means computing `includes_for_smart_search` in the extended branch too, which changes the JOIN set for every collection using dotted `search_fields`.

---

**`Preferential` — `query_for_batch` has the same missing-join hole. Correct, deliberately not fixed here.**

`@base_records_for_batch = @records` is captured before anything eager-loads (`resources_getter.rb:19`), and `get_ids_from_request` → `fetch_ids` → `find_in_batches` calls neither `perform` nor `optimized_count`. A bulk smart action over "select all records" with extended search on hits the same 500.

Same reasoning: pre-existing, a different route (bulk actions, not browse), and it needs its own test surface — none of the specs here exercise `get_ids_from_request`. Its own ticket.

---

Neither `Preferential` item was reproduced — both are static analysis, stated as such in the review. I would not open tickets off unexecuted findings without confirming them first.

matthv:
## Review

Solid fix — the diagnosis is right and the `count == list.length` + absolute-value double assertion in the new specs is good discipline. Two things worth a look before merge:

### 1. `resources_getter.rb:16` — `ResourcesGetter` doesn't get the same protection as `HasManyGetter`

```ruby
@search_query_builder = SearchQueryBuilder.new(@params, @includes, @collection, @user)
```

`@includes` here is still unfiltered. Under `searchExtended=1`, `compute_includes` (resources_getter.rb:126) only rejects **cross-database** associations before handing `@includes` to `SearchQueryBuilder` — it doesn't subtract instance-dependent-scoped associations, unlike `perform`/`optimized_count` in this same file, which route through the full `analyze_associations` (polymorphic + cross-db + instance-dependent) before eager-loading.

So a `belongs_to`/`has_one` with an instance-dependent scope (e.g. `Tree#eponymous_island` in the dummy app) can still end up in a `WHERE` predicate for a table `optimize_record_loading` never joins — the exact "missing FROM-clause entry" 500 this PR sets out to fix, just reachable via `ResourcesGetter` instead of `HasManyGetter`. It's masked in the current specs only because `eponymous_island`'s table (`islands`) happens to already be joined via the plain `:island` association.

Suggest hoisting `HasManyGetter#searchable_includes` into `BaseGetter` and having both getters call it, so the join set and the search-builder set can't drift apart again (this PR's own framing — "a single `@includes` was serving three different jobs" — applies here too).

### 2. Two new specs assert an association is eager-loaded when it isn't

`resources_getter_spec.rb:667` and `:682` (`'eager loads every one-association the search predicates reach'` / `'eager loads the same associations as the request carrying no fields'`) assert `getter.includes` contains `:eponymous_island`. That association is routed into `preload_loads` by `analyze_associations`, so it's never eager-loaded (and with `force_preload: false` in `perform`, isn't loaded at all here). The assertion is really pinning "what gets passed to the search builder," not "what's eager-loaded" — i.e. it pins point 1's gap as expected behavior without surfacing it.

Suggest renaming (e.g. "passes every one-association to the search builder") or, once point 1 is fixed, asserting against the actually-joined set so a future regression here would fail loudly.

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): Both points were right. Fixed in 9e553f8.

## 1. Broader than it looks from the diff

You framed this as masked by the dummy app. It is worse than that — it is reachable with the models already in the repo. `Island`'s one-associations are `location` (table `locations`) and the instance-dependent `eponymous_tree` (table `trees`), and nothing else joins `trees`:

```
one_associations(Island) = [:location, :eponymous_tree]
GET /forest/Island        → SQLite3::SQLException: no such column: trees.name
GET /forest/Island/count  → SQLite3::SQLException: no such column: trees.name
```

Both routes, on the branch, before this commit. `Tree` hides it only because `eponymous_island` resolves against the `isle` join that `island` already contributes — exactly as you said, but the consequence is a whole collection 500ing on its own list, not just a latent asymmetry.

That also means the regression test was cheap, which changed the call. I had been leaning toward a follow-up ticket on the assumption it needed new dummy models; `Island` already is the failing shape.

`searchable_includes` now lives on `BaseGetter` as you suggested, taking the resource as a parameter — it cannot be a straight hoist, since `HasManyGetter` passes `model_association` and `ResourcesGetter` passes `@resource`. Net production change is +8/-10/+1; `has_many_getter.rb` got shorter.

Two specs added, red before the fix and mutation-checked: reverting only the `ResourcesGetter` call site fails exactly those two. One pins that the un-joined association no longer breaks the query; the other pins that a search reaching `location` still matches, so a future over-narrowing fails loudly rather than silently returning fewer rows.

## 2. Renamed, though not to your wording

Agreed, and it is a straight violation of `testing.md#Test name states the exact behavior it asserts` — the assertion was accurate about `@includes`, the name was not.

I did not use "passes every one-association to the search builder", because after point 1 that is false: `eponymous_island` is no longer passed. They now read `keeps every one-association in the include set when no fields are requested` / `keeps the same include set when fields are requested`, which is the projection-independence property they actually pin. `@includes` is still the eager-load candidate set; only what reaches the search builder narrowed.

## Scope note — this narrows behaviour

Worth being explicit, because it is not purely a bug fix.

Extended search no longer matches through an instance-dependent one-association. Today, in an app where such an association's table happens to be joined by a sibling, that match works and silently disappears with this change. `Tree` is that case in the dummy app: `eponymous_island` contributed a duplicate predicate on `isle`, so nothing observable moves there — but a real app with a distinct table would see extended search stop matching through it.

The alternative is a 500, so I think the trade is clear. It is a behaviour change rather than a pure fix, and it belongs in the release note.

Verified after the change:

| `searchExtended=1` on `Island` | before | after |
| -- | -- | -- |
| `search=skull` (own column) | raised | `count=1`, list 1 |
| `search=12345` (through `location`) | raised | `count=1`, list 1 |

488 examples, 0 failures; `rake test` green; CI green on 9e553f8.

app/services/forest_liana/has_many_getter.rb
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Correction to my reply above — the last paragraph is now out of date.

I wrote that `ResourcesGetter#compute_includes` carries the same latent hole and that I was flagging it rather than folding it in. @matthv raised the same gap in review, and it turned out to be reachable with models already in the repo rather than latent: `Island`'s one-associations are `location` and the instance-dependent `eponymous_tree`, nothing else joins `trees`, so `GET /forest/Island` was returning 500 on both its list and its count.

Fixed in 9e553f8. `searchable_includes` now lives on `BaseGetter` and both getters use it.

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): Heads-up on a collision with #790, which landed on `main` while this was in review. @matthv — you co-authored it, so this is mostly for you.

**There is a 500 on `main` today**, at 9.20.11, independent of this branch:

```
HasManyGetter, searchExtended=1
  search='kiwi'  → count=2 records=2
  search=''      → ActiveRecord::StatementInvalid: no such column: users.name
```

#790 added `associations_referenced_by_extended_search`, which gates on `@params[:search].present?`. An empty search string is not `present?`, so it returns `[]`, the one-associations move from `eager_load` to `preload`, and their tables are never joined. But `SearchQueryBuilder#search_param` guards on `if @search`, and `''` is truthy in Ruby — so it still emits `LOWER(users.name) LIKE '%%'`. Predicate on an unjoined table.

This is the same `.present?` vs `.nil?` distinction this PR already hit from the other side: `ResourcesGetter` derives `@count_needs_includes` from `!@params[:search].nil?` precisely because `search=''` still builds `LIKE '%%'` predicates, and there is a spec pinning it. `.present?` there would have reintroduced the original bug.

**To be clear about attribution:** this branch neither causes that 500 nor fixes it. I reproduced it on `a8fec93` with this branch nowhere in the picture, and again on a local merge of `main` into this branch — same failure either way.

**The structural point.** #790 and this PR now solve the same coupling from opposite directions, in the same class:

* `associations_referenced_by_extended_search` (#790) asks *"which associations will the search reference, so keep their JOIN?"*
* `searchable_includes` (this PR, now on `BaseGetter`) asks *"which associations are actually joined, so the search may reference them?"*

They agree whenever a non-empty search is present, and diverge on the empty string. After a merge, `searchable_includes` computes from `analyze_associations` and so over-reports relative to #790's narrower `keep_eager` — it will hand the builder associations that #790 has moved to preload. Two sources of truth for one invariant is exactly the drift you warned about in your review of this PR, and it has already produced one live bug.

**Suggestion,** in preference order:

1. Change #790's gate to `!@params[:search].nil?` so it matches `SearchQueryBuilder`'s own `if @search`. One-word fix, closes the `main` bug.
2. Then reconcile the two: have `associations_to_keep_eager` intersect with `searchable_includes`, so the joined set and the searchable set cannot disagree by construction.

Happy to do either here or in a separate PR against `main` — the `main` bug is arguably worth its own fix since it ships in 9.20.11 regardless of what happens to this branch. Tell me which you prefer and I'll pick it up.

I have not merged `main` into this branch; it is still on `9e553f8`.

matthv:
Sounds good on both counts — thanks for the fast turnaround and the thorough writeup on the `Island` repro, that's a much better argument than mine for why this needed fixing now rather than later.

On the `main`/9.20.11 regression: agreed it's ours to fix, not yours. We'll open a separate PR against `main` for it (option 1, the one-word gate fix, then reconcile with `searchable_includes` per your suggestion) rather than block this PR on it.

We'll approve this one once CI is green on `9e553f8`.

matthv (approved):

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): Thanks — confirming CI is green on `9e553f8` (Tests, Linting, Macroscope all passing; the 7 annotations are the pre-existing Node 20 / `set-output` / coverage-path warnings, unrelated to this branch).

Standing down on the `main` regression so we do not both write it — it is yours. The repro if it is useful: `HasManyGetter` with `searchExtended=1` and `search=''`, on any related collection whose target has a searchable one-association. `spec/dummy` reproduces it with `Island` → `trees` and an owner named anything.

One thing worth carrying into the release note for whichever of us ships first: this branch **narrows behaviour** as well as fixing the 500 — extended search no longer matches through an instance-dependent one-association. Where such an association's table happened to be joined by a sibling, that match works today and stops. `Tree` is exactly that case in the dummy app, which is why no existing spec moved.

Not merging — that is Brian's call.

matthv:
CI confirmed green, thanks. And appreciated on standing down cleanly — repro was exactly what we needed.

Opened #792 for the \`main\` regression: one-line gate fix (\`.present?\` → \`!@params[:search].nil?\`), same reasoning \`ResourcesGetter\` already applies, regression spec included (red before the fix, confirmed by reverting locally).

Noted your point on the behavior narrowing (extended search no longer matching through an instance-dependent one-association) for the release note once this one ships.

PR #792, opened 2026-08-25
fix(has-many-getter): gate extended search on nil, not present, search

matthv:
## What breaks today (main / 9.20.11)

With extended search on (`searchExtended=1`) and an **empty** search string, a has-many related list 500s:

```
GET /forest/Island/:id/relationships/trees?search=&searchExtended=1
→ ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: users.name
```

## First fix: gate on `nil`, not `present?`

`associations_referenced_by_extended_search` (introduced in #790) decided whether a display-only association must stay eager-loaded, gated on `@params[:search].present?`.

`SearchQueryBuilder#search_param` guards on `if @search` instead, and `''` is truthy in Ruby — so an empty search string still built `LIKE '%%'` predicates against every included one-association's table. `.present?` is `false` for `''`, so the association moved to `preload` (no JOIN) while the predicate still ran against it: missing FROM-clause entry.

`ResourcesGetter` already hit this exact distinction and uses `!@params[:search].nil?` for the same reason. First commit aligned `HasManyGetter`'s gate with it.

## Second fix: the actual root cause — `''` should be a no-op, not a filter

Review on this PR flagged that gating on presence just patches the symptom: with the first fix alone, `search=''` stopped 500ing but became a real *filter* — `LOWER(NULL) LIKE '%%'` evaluates to `NULL` in SQL, so a record whose searched columns are all `NULL` is silently excluded under `search=''` while present with no `search` param at all.

Root cause: `SearchQueryBuilder#initialize` kept `@search = @params[:search]` as-is, so every `if @search` check downstream treated `''` as truthy. Fixed at the source with `.presence`, so `''` behaves like `nil` everywhere.

This retires both `!nil?`-instead-of-`.present?` workarounds now that `''` no longer needs the extra JOIN/eager-load:
- `has_many_getter.rb#associations_referenced_by_extended_search` → back to `.present?`
- `resources_getter.rb`'s `@count_needs_includes` (from #791) → back to `.present?`

## Tests

- Regression spec for the original 500 (has_many_getter_pagination_spec.rb), red before the first fix.
- New spec: a record with `name: nil` and no owner is present under extended search with no `search` param, and stays present under `search=''` — red before the second fix (confirmed by reverting locally).

Full suite: 498 examples, 0 failures.

> [!NOTE]
> ## What breaks today (main / 9.20.11)
>
> With extended search on (`searchExtended=1`) and an **empty** search string, a has-many related list 500s:
>
> ```
> GET /forest/Island/:id/relationships/trees?search=&searchExtended=1
> → ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: users.name
> ```
>
> ## First fix: gate on `nil`, not `present?`
>
> `associations_referenced_by_extended_search` (introduced in #790) decided whether a display-only association must stay eager-loaded, gated on `@params[:search].present?`.
>
> `SearchQueryBuilder#search_param` guards on `if @search` instead, and `''` is truthy in Ruby — so an empty search string still built `LIKE '%%'` predicates against every included one-association's table. `.present?` is `false` for `''`, so the association moved to `preload` (no JOIN) while the predicate still ran against it: missing FROM-clause entry.
>
> `ResourcesGetter` already hit this exact distinction and uses `!@params[:search].nil?` for the same reason. First commit aligned `HasManyGetter`'s gate with it.
>
> ## Second fix: the actual root cause — `''` should be a no-op, not a filter
>
> Review on this PR flagged that gating on presence just patches the symptom: with the first fix alone, `search=''` stopped 500ing but became a real *filter* — `LOWER(NULL) LIKE '%%'` evaluates to `NULL` in SQL, so a record whose searched columns are all `NULL` is silently excluded under `search=''` while present with no `search` param at all.
>
> Root cause: `SearchQueryBuilder#initialize` kept `@search = @params[:search]` as-is, so every `if @search` check downstream treated `''` as truthy. Fixed at the source with `.presence`, so `''` behaves like `nil` everywhere.
>
> This retires both `!nil?`-instead-of-`.present?` workarounds now that `''` no longer needs the extra JOIN/eager-load:
> - `has_many_getter.rb#associations_referenced_by_extended_search` → back to `.present?`
> - `resources_getter.rb`'s `@count_needs_includes` (from #791) → back to `.present?`
>
> ## Tests
>
> - Regression spec for the original 500 (has_many_getter_pagination_spec.rb), red before the first fix.
> - New spec: a record with `name: nil` and no owner is present under extended search with no `search` param, and stays present under `search=''` — red before the second fix (confirmed by reverting locally).
>
> Full suite: 498 examples, 0 failures.
> #### Changes since #792 opened
>
> - Renamed RSpec example description for `ForestLiana::HasManyGetter` pagination spec and added count assertion [045a126]
> 
>

### DIFF
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -183,8 +183,11 @@ module ForestLiana
     # every included one-association's columns and relies on its table already being joined.
     # Mirror the associations it can reach so their JOIN stays eager instead of moving to
     # preload (which never joins, on any Rails version).
+    #
+    # NOTICE: `!nil?`, not `.present?` — search_param guards on `if @search`, and `''` is
+    # truthy, so an empty search string still needs the JOIN.
     def associations_referenced_by_extended_search
-      return [] unless @params[:search].present? && @params['searchExtended'].to_i == 1
+      return [] unless !@params[:search].nil? && @params['searchExtended'].to_i == 1
 
       target_model = model_association
       includes_symbols = @includes.map(&:to_sym)
--- a/app/services/forest_liana/has_many_getter.rb
+++ b/app/services/forest_liana/has_many_getter.rb
@@ -194,11 +194,8 @@ module ForestLiana
     # every included one-association's columns and relies on its table already being joined.
     # Mirror the associations it can reach so their JOIN stays eager instead of moving to
     # preload (which never joins, on any Rails version).
-    #
-    # NOTICE: `!nil?`, not `.present?` — search_param guards on `if @search`, and `''` is
-    # truthy, so an empty search string still needs the JOIN.
     def associations_referenced_by_extended_search
-      return [] unless !@params[:search].nil? && @params['searchExtended'].to_i == 1
+      return [] unless @params[:search].present? && @params['searchExtended'].to_i == 1
 
       target_model = model_association
       includes_symbols = @includes.map(&:to_sym)
--- a/spec/services/forest_liana/has_many_getter_pagination_spec.rb
+++ b/spec/services/forest_liana/has_many_getter_pagination_spec.rb
@@ -158,5 +158,25 @@ module ForestLiana
         expect(records.map(&:name)).to contain_exactly('cedar', 'pine')
       end
     end
+
+    describe 'with extended search and an empty search string' do
+      let(:params) do
+        {
+          id: island.id,
+          association_name: 'trees',
+          search: '',
+          'searchExtended' => '1',
+          page: { size: 15, number: 1 },
+          timezone: 'America/Nome'
+        }
+      end
+
+      it 'keeps the display association joined instead of raising on the un-joined table' do
+        subject.perform
+
+        expect { subject.records.to_a }.not_to raise_error
+        expect(subject.records.to_a.map(&:name)).to contain_exactly('cedar', 'olive', 'pine', 'fig')
+      end
+    end
   end
 end
--- a/spec/services/forest_liana/has_many_getter_pagination_spec.rb
+++ b/spec/services/forest_liana/has_many_getter_pagination_spec.rb
@@ -171,11 +171,12 @@ module ForestLiana
         }
       end
 
-      it 'keeps the display association joined instead of raising on the un-joined table' do
+      it 'treats an empty search as a no-op instead of building predicates it cannot join' do
         subject.perform
 
         expect { subject.records.to_a }.not_to raise_error
         expect(subject.records.to_a.map(&:name)).to contain_exactly('cedar', 'olive', 'pine', 'fig')
+        expect(subject.count).to eq(@trees.size)
       end
 
       describe 'with a record whose searched columns are all NULL' do
--- a/spec/services/forest_liana/has_many_getter_pagination_spec.rb
+++ b/spec/services/forest_liana/has_many_getter_pagination_spec.rb
@@ -177,6 +177,16 @@ module ForestLiana
         expect { subject.records.to_a }.not_to raise_error
         expect(subject.records.to_a.map(&:name)).to contain_exactly('cedar', 'olive', 'pine', 'fig')
       end
+
+      describe 'with a record whose searched columns are all NULL' do
+        before(:each) { Tree.create!(name: nil, island: island) }
+
+        it 'is a no-op, not a filter: the record is still returned' do
+          subject.perform
+
+          expect(subject.records.to_a.map(&:name)).to contain_exactly('cedar', 'olive', 'pine', 'fig', nil)
+        end
+      end
     end
   end
 end
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -6,7 +6,7 @@ module ForestLiana
       @resource = resource
       @params = params
       @user = forest_user
-      @count_needs_includes = !@params[:search].nil?
+      @count_needs_includes = @params[:search].present?
       @collection_name = ForestLiana.name_for(@resource)
       @collection = get_collection(@collection_name)
       @fields_to_serialize = get_fields_to_serialize
--- a/app/services/forest_liana/search_query_builder.rb
+++ b/app/services/forest_liana/search_query_builder.rb
@@ -9,7 +9,8 @@ module ForestLiana
       @includes = includes
       @collection = collection
       @fields_searched = []
-      @search = @params[:search]
+      # '' is truthy, so without .presence an empty search still builds LIKE '%%' predicates.
+      @search = @params[:search].presence
       @user = user
     end
 


app/services/forest_liana/has_many_getter.rb
```diff
@@ -183,8 +183,11 @@ def associations_referenced_by_sort
     # every included one-association's columns and relies on its table already being joined.
     # Mirror the associations it can reach so their JOIN stays eager instead of moving to
     # preload (which never joins, on any Rails version).
+    #
+    # NOTICE: `!nil?`, not `.present?` — search_param guards on `if @search`, and `''` is
+    # truthy, so an empty search string still needs the JOIN.
     def associations_referenced_by_extended_search
-      return [] unless @params[:search].present? && @params['searchExtended'].to_i == 1
+      return [] unless !@params[:search].nil? && @params['searchExtended'].to_i == 1
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): `Should fix`

An operator with extended search on and an **empty** search box now gets a silently short page instead of a 500 — the records are dropped rather than reported.

Reproduced on this branch merged with current `main`, using this file's own fixture (4 trees, the mis-declared `has_one :island_neighbour` that resolves to 4 rows each):

| `page[size]=2` | rows returned | count |
| -- | -- | -- |
| no search | `["cedar","olive"]` | 4 |
| `search=''` + `searchExtended=1` | **`["cedar"]`** | 4 |

That is exactly the row-multiplication truncation #790 was written to prevent. `associations_referenced_by_extended_search` returns *every* one-association in `@includes`, and `associations_to_keep_eager` puts them all back into `eager_load` — the set #790 had just moved to `preload`. `LIMIT` then applies to the multiplied JOIN output.

**This is not introduced by this PR, and the framing matters.** I checked the same request with the gate reverted to `.present?` — i.e. `main` as it ships today — and got the identical `rows=1, count=4`. The truncation is already live for every **non-blank** extended search. What this PR does is route `search=''` and `search=' '` into it as well, and those previously 500'd. So it trades a loud error for silent data loss on two more inputs.

The fix itself is right; the gap is that it hands back JOINs the search does not need. A cheap tightening in the same method: when `@collection.search_fields` is set and names no `assoc.column`, `SearchQueryBuilder` emits no association predicate at all (`search_query_builder.rb`, the `@collection.search_fields.nil? || association_search.include?(column.name)` condition) — yet this method still forces every JOIN. Mirroring `search_fields` here would shrink the blast radius without touching #790.

Worth deciding deliberately rather than inheriting it.

spec/services/forest_liana/has_many_getter_pagination_spec.rb
```diff
@@ -158,5 +158,25 @@ module ForestLiana
...
+          id: island.id,
+          association_name: 'trees',
+          search: '',
+          'searchExtended' => '1',
+          page: { size: 15, number: 1 },
+          timezone: 'America/Nome'
+        }
+      end
+
+      it 'keeps the display association joined instead of raising on the un-joined table' do
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): `Violates conventions` — `testing.md#Cover error and edge paths, not only the happy path`

Two gaps in the new case, both of which let the finding above pass unnoticed.

**1. `page[size]: 15` hides the truncation the file exists to catch.** With the fix, this example's relation eager-loads `island_neighbour`, so the 4 trees produce 16 joined rows against a `LIMIT 15`. It passes only because ActiveRecord dedupes by primary key and no single tree loses all four of its rows. Drop to the file's own default `page_size` of 2 and the same request returns **1 tree, not 4** — I ran it. A regression spec in a file titled *"pagination with a has_one that is one-to-many in the data"* is currently structurally blind to that.

**2. `#count` is never asserted**, and it is a separate caller-reachable entry point that breaks the same way. `AssociationsController#count` (`associations_controller.rb:27-44`) builds a `HasManyGetter` from the same params and calls `getter.count`, which issues different SQL — `COUNT(DISTINCT …)`, plus the composite-PK adapter branches at `has_many_getter.rb:28-50`. Verified: with the fix `count = 4`; with the gate reverted it **raises** `no such column: users.name`. So the count endpoint is half of what this PR fixes, and it is the untested half.

This file already has the idiom at `:91-95` — `expect(subject.count).to eq(@trees.size)`. Adding that line plus a smaller `page[size]` turns this into a test that would fail on both defects.

The `not_to raise_error` half is sound and does its job; this is about the assertions around it.

hercemer42 (commented):
**Spec:** no Linear ticket governs this PR — no `linear-code[bot]` linkback, no `PRD-` id in the body or branch (`fix/extended-search-empty-string-gate`), and the commit footer references a GitHub discussion rather than a ticket. The functional check below is therefore against the PR description only. For the record, PRD-1034 does **not** cover this: it closed at 09:29 today and its own test-pass note explicitly disclaims this defect as "being fixed separately by the #790 authors". Worth a ticket if this line of work should be trackable — #790 has none either.

Against the description, the fix does what it claims. I verified it on this branch merged with current `main` (`4ebea16`, which now carries #791): full suite **497 examples, 0 failures**, and the new spec is a genuine regression test — reverting the one-line gate turns it red with exactly `no such column: users.name`. The `!nil?` / `if @search` alignment is correct, and the two mechanisms that now coexist in this class (`associations_referenced_by_extended_search` and #791's `searchable_includes`) resolve to identical sets once the gate matches, so the merge is safe.

Two findings inline. One note below.

---

Claude Opus 5 (claude-opus-5[1m]): `Preferential`

**Applies to:** `app/services/forest_liana/has_many_getter.rb:190` and `search_query_builder.rb#search_param`

An empty search string is now a *filter* on related lists rather than a no-op, and that drops rows. `search=''` builds `LIKE '%%'` over every searched text column, and `LOWER(NULL) LIKE '%%'` evaluates to NULL — so a record whose searched text columns are all NULL is excluded from both the list and the count. Concretely: a `Tree` with a NULL `name` and no owner is present under `?searchExtended=1` with no `search` key, and absent under `?search=&searchExtended=1`.

Not a regression, and arguably the right call — #791 already committed the top-level path to the same semantic (`resources_getter.rb`, `!@params[:search].nil?`), and this makes the related-list path consistent with it. Flagging only so it is a deliberate choice. The alternative — guarding `search_param` on `@search.present?` instead — would remove the 500, the unnecessary JOINs and this NULL-drop in one line, at the cost of changing top-level behaviour too.

matthv:
Went with your alternative — fixed at the source instead of gating around it.

`SearchQueryBuilder#initialize` now normalizes `@search` via `.presence`, so `''` is a true no-op everywhere `if @search` is checked (no more `LIKE '%%'`, no more NULL-drop). That let us retire both `!nil?`-instead-of-`.present?` workarounds now that `''` no longer needs the extra JOIN/eager-load: `resources_getter.rb`'s `@count_needs_includes` and `has_many_getter.rb`'s `associations_referenced_by_extended_search` both go back to `.present?`.

Added a regression spec for the NULL case (a record with `name: nil` and no owner, present with no `search` param, still present under `search=''`) — red before this commit. Full suite: 498 examples, 0 failures.

app/services/forest_liana/has_many_getter.rb
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Resolved by c035957 — and the source fix is better than the tightening I suggested. Verified on `c035957`, which is now level with `main`.

Normalizing `@search` with `.presence` means an empty search never reaches `search_param`, so no association predicate is emitted and there is nothing to keep joined. The request stops entering the row-multiplication path rather than entering it more carefully:

| `page[size]=2` | rows | count | LEFT OUTER JOINs |
| -- | -- | -- | -- |
| no search | `["cedar","olive"]` | 4 | none |
| `search=''` + `searchExtended=1` | `["cedar","olive"]` | 4 | **none** |
| `search='e'` + `searchExtended=1` | `["cedar"]` | 4 | users, isle, locations, trees |

That middle row returned `["cedar"]` before this commit. Both of my findings on this line are closed by it — the truncation on blank searches, and the `Preferential` note about `''` filtering out records whose searched columns are all NULL. You took the alternative and pinned it with a spec.

The third row is the pre-existing #790 truncation on non-blank extended search, untouched here and out of scope as discussed.

Two things I checked specifically because this commit reverts a line #791 shipped deliberately (`resources_getter.rb`'s `!@params[:search].nil?`). Both clean:

* **No consumer is left out of sync.** `params[:search]` has exactly three readers — `resources_getter.rb:9`, `has_many_getter.rb:198`, `search_query_builder.rb:13` — now consistently `.present?` / `.presence`. Both `SearchQueryBuilder` call sites come from getters using the same predicate, so the guard and the normalization cannot drift apart. That was the failure mode behind this whole line of bugs, so it is worth saying explicitly that it is now closed by construction.
* **#791's specs still hold.** All 63 examples across `count_spec.rb` and `resources_getter_spec.rb` pass. The empty-search count assertion still returns 3 — now because no predicate is built, rather than because `LIKE '%%'` matched everything. Same answer, sounder mechanism.

Every `@search` use inside the builder sits behind `if @search` or the `is_a?(String)` guard in `malformed_uuid_search?`, so nil is safe throughout. Full suite 498 examples, 0 failures, and both new specs go red when `.presence` is reverted.

spec/services/forest_liana/has_many_getter_pagination_spec.rb
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Half of this is now moot, half still stands, and the redesign has introduced a third thing. Leaving the thread open for the last two — neither is worth holding the PR on.

**Moot: the `page[size]: 15` blindness.** That mattered only because the empty search forced every one-association back into `eager_load`. It no longer does — I measured `LEFT OUTER JOIN` count as zero for this request on `c035957`. Four trees, no multiplication, nothing for `LIMIT 15` to truncate. Disregard that half.

**New: the test name is now false.** `'keeps the display association joined instead of raising on the un-joined table'` describes the old design. Under `.presence` nothing is joined — the search is a no-op, and the example passes because no predicate is built, not because the association survived in `eager_load`. Verified: `joins=[]` for exactly these params.

It is still a genuine regression test — reverting `.presence` turns it red with `no such column: users.name` — so only the name is wrong, and `testing.md#Test name states the exact behavior it asserts` is explicit that the name has to move when the assertion's meaning does. Something like `'treats an empty search as a no-op instead of building predicates it cannot join'` would match what it now proves. The nested NULL example's name is already accurate.

**Still stands: `#count` is untested.** `AssociationsController#count` builds a `HasManyGetter` from the same params and calls `getter.count`, which issues different SQL — `COUNT(DISTINCT …)` plus the composite-PK adapter branches at `has_many_getter.rb:28-50`. It breaks the same way pre-fix: with `.presence` reverted, `count` raises `no such column: users.name`, same as `records`. So the count route is half of what this PR fixes and is the half no example touches.

This file already has the idiom at `:91-95` — `expect(subject.count).to eq(@trees.size)`. One line in either example covers it.

Your call on both; I would not block on either.

hercemer42 (dismissed):

spec/services/forest_liana/has_many_getter_pagination_spec.rb
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Both addressed in 045a126 — resolving.

Verified at that SHA rather than read off the diff:

* **The rename is accurate.** `'treats an empty search as a no-op instead of building predicates it cannot join'` is what the example now proves — I measured zero `LEFT OUTER JOIN`s for exactly these params, so there is no association being kept joined and the name no longer describes the pre-`.presence` design.
* **The count assertion is meaningful, not vacuous.** `subject.count` returns 4 against a 4-tree fixture, so `eq(@trees.size)` is pinning a real value rather than passing trivially. `AssociationsController#count` now has coverage on the path that broke.
* **Still a genuine regression test.** Reverting `.presence` puts 2 failures back in this block.

Full suite 498 examples, 0 failures.

That closes every finding from my review. For the record, the one thing this PR does *not* close is the pre-existing #790 truncation on non-blank extended search — `search='e'` + `searchExtended=1` still returns 1 row against a count of 4 on `main`. Out of scope here and yours to schedule; the reproduction is in the resolved thread above if it is useful later.

hercemer42 (approved):

PR #796, opened 2026-09-02
feat(permissions): refuse a filter or sort reading a collection the role cannot read

matthv:
fixes PRD-1084

Depends on #794 (PRD-1081) and #795 (PRD-1083, `redact_fields`/the "named vs default" pattern this ticket mirrors for refusal instead of redaction) — branched from #795, not mergeable on its own until those two land; the diff below includes their commits until then.

## Why

Redacting the response (#795) is not enough: a filter on a column of a collection the role cannot read leaks its value without ever returning it — a `starts_with` answers one guess per request, read off whether rows come back — and a sort on such a column silently reorders the page by it. Both leak the value they touch, so this guard always **refuses**, never redacts: dropping a filter condition widens the result set and dropping a sort clause silently reorders it.

## What

`Ability::Permission#assert_can_read_query_fields` resolves each filter/sort path to its leaf collection (`FieldPath`, #794) and refuses on the first unreadable one, naming the path and the collection. The root collection is pinned readable — already gated upstream by `browse`/`read` — and the agent-injected scope is never checked, only the caller's own filter tree.

The two are wired differently, following where each is actually parsed:

* **Filters** are checked inside `SearchQueryBuilder#perform`, the single funnel both getters build `@records` through (list, count, csv, related, and destroy_bulk/select-all) — so filters are refused on every one of those routes alike, count included.
* **Sort** turned out to be parsed inside the getters' own *constructors* (`prepare_query`), identically whether the caller asked for `index` or `count` — there is no perform/count branch to hook at that layer. `sort_query` now only records the paths it resolves; a separate `assert_sort_readable!` checks them, called explicitly by `ResourcesGetter#perform`/`HasManyGetter#perform` and deliberately never by `count`, which never applies the sort anyway (Rails strips the ORDER BY from the emitted COUNT SQL on its own).

`ScopeManager.append_scope_for_user` is split into `append_scope` (merges an already-injected filter with the scope) so the caller-only tree can be read once, checked, and handed on unchanged — the guard authorizes the exact object `FiltersParser` applies, not a second, independent parse that could drift from it. Calling `append_scope_for_user` a second time on an already-injected tree isn't safe: it re-runs the `{{...}}` placeholder substitution, which raises on a value that merely looks like one.

`UnauthorizedQueryFieldError` (403) is wired into the four routes that build a filtered/sorted query: index, count, and their relationships equivalents. **`associations#count` had no `ExpectedError` rescue at all before this** — any denial surfaced as a bodiless 500; it now matches the other three.

A path `FieldPath` cannot resolve is treated differently by design: a filter's is left to raise from `FiltersParser` right after this guard, which already has its own pinned 422 message for it; a sort's has no such downstream validator, so it is left to raise directly from `FieldPath` instead of silently letting a bogus sort through to a 500.

## A pre-existing test-isolation gap this surfaced

Three spec files (`resources_getter_spec.rb`, `has_many_getter_pagination_spec.rb`, and the scoped-action examples in `actions_controller_spec.rb`) construct getters directly and never anticipated any permission-system check running during that — nothing before this guard read `has_permission_system?` outside of `forest_authorize!`, which those specs bypass. The repo's `Rails.cache` is a `FileStore` persisted on disk across separate `bundle exec rspec` invocations, so these files intermittently inherited stale permission state from other spec runs. They now declare "no permission system" explicitly — the same fix `count_spec.rb` already needed for the same reason in #795.

## Not in scope

* The three chart stat-getters share the same `append_scope_for_user` seam but are not covered here, per the ticket.
* `destroy_bulk`/select-all now goes through this guard with no exemption — a deliberate divergence from v2, which carved delete routes out (PRD-1012); here the filter comes from a list view the guard has already authorized once, and `forest_authorize!('delete')` gates the route itself.

## Behavioral change to flag for release notes

`AssociationsController#index`/`#count` now call `forest_authorize!('browse', ...)` on the related collection — previously neither route checked any permission at all (the actual security gap this PR also closes for `update`/`associate`/`dissociate`, added while addressing review feedback). A role that could browse a parent record and see a related list without `browse` on the related collection itself will now get a 403 there. This is the correct target state, but it's a visible behavior change for any existing deployment relying on the old (unauthorized) default.

## Tests

New: `spec/services/forest_liana/ability/assert_can_read_query_fields_spec.rb` (13 unit examples — denial/allow on filter and sort, traversal through an unreadable collection, root pin, polymorphic ×3, smart belongsTo, unresolvable-path asymmetry), `FiltersParser.field_paths` specs, `SearchQueryBuilder#assert_sort_readable!`/tree-fidelity specs, plus new request-level examples in `resources_spec.rb` and `associations_spec.rb` covering index/count/csv, the scope-not-checked and root-pinned cases.

Full suite: 588 examples, 0 failures. Root pin, count-never-checks-sort, and the filter-path-skip-vs-sort-path-raise asymmetry are each mutation-checked (reverting the corresponding line turns the relevant spec red).

> [!NOTE]
> ### Refuse filters and sorts reading unreadable collections in permission checks
> - Adds `ForestLiana::Ability::Permission#assert_can_read_query_fields`, which resolves filter and sort field paths to their target collections and raises `UnauthorizedQueryFieldError` (HTTP 403) when any leaf collection is not readable
> - Wires the check into `SearchQueryBuilder#perform` for caller filters and into `assert_sort_readable!` on `ResourcesGetter` and `HasManyGetter` so index, count, CSV, and association listing all enforce it
> - Adds association write authorization in `AssociationsController`: edit permission on the write target for associate/update, and delete permission when dissociation or has_one replacement destroys records (`BelongsToUpdater.replaces_destructively?`, `HasManyDissociator.destroys_on_unlink?`)
> - Extracts scope composition into `ScopeManager.append_scope` so the caller filter is context-injected once before scope appending
> - Fixes malformed aggregator handling: non-Array `conditions` values in `FiltersParser` and `ContextVariablesInjector` now raise HTTP 422 instead of iterating the invalid shape
> - Risk: requests that previously sorted or filtered through an unreadable related collection now return 403; `ResourcesGetter.initialize_resources_getter` applies the sort check during construction so `get_ids_from_request` paths are also covered — verify no internal callers depend on sorting through unreadable relations
>
> 
> #### Changes since #796 opened
>
> - Added request-level validation to reject non-Hash top-level filters with a 422 status code and error message 'Filters cannot be a raw value' [8735ad3]
> - Added type guards in `ForestLiana::FiltersParser.field_paths` and `ForestLiana::Utils::ContextVariablesInjector.inject_context_in_filter` to handle non-Hash filter inputs [8735ad3]
> - Added test coverage for non-Hash filter handling in `FiltersParser.field_paths` and resource requests [8735ad3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b626556.</sup>
> 
>

### DIFF
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -123,6 +123,13 @@ module ForestLiana
         head :no_content
       rescue ActiveRecord::RecordNotDestroyed => error
         render json: { errors: [{ status: :bad_request, detail: error.message }] }, status: :bad_request
+      rescue ForestLiana::Errors::ExpectedError => error
+        error.display_error
+        error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
+          status: error.error_code,
+          detail: error.message
+        }])
+        render(serializer: nil, json: error_data, status: error.status)
       rescue => error
         FOREST_REPORTER.report error
         FOREST_LOGGER.error "Association Dissociate error: #{error}\n#{format_stacktrace(error)}"
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -15,10 +15,13 @@ module ForestLiana
     end
 
     def index
+      # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
+      # Authorized outside the begin, like update/associate/dissociate, so a denial reaches
+      # ApplicationController's rescue_from and keeps its name/data instead of falling into the
+      # generic ExpectedError rescue below.
+      action = request.format == 'csv' ? 'export' : 'browse'
+      forest_authorize!(action, forest_user, @association.klass)
       begin
-        # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
-        action = request.format == 'csv' ? 'export' : 'browse'
-        forest_authorize!(action, forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.perform
 
@@ -60,8 +63,9 @@ module ForestLiana
       #         through and dereferencing a nil @association, which surfaced as
       #         a double-render / 500 rather than the intended 404.
       return if performed?
+      # Authorized outside the begin, like index above, so a denial keeps its name/data.
+      forest_authorize!('browse', forest_user, @association.klass)
       begin
-        forest_authorize!('browse', forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.count
 
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -94,6 +94,7 @@ module ForestLiana
         # for a belongsTo.
         edit_subject = @association.macro == :has_one ? @association.klass : @resource
         forest_authorize!('edit', forest_user, edit_subject)
+        forest_authorize!('delete', forest_user, @association.klass) if BelongsToUpdater.replaces_destructively?(@association)
         updater = BelongsToUpdater.new(@resource, @association, params)
         updater.perform
 
@@ -119,7 +120,7 @@ module ForestLiana
 
     def associate
       begin
-        forest_authorize!('edit', forest_user, @association.klass)
+        forest_authorize!('edit', forest_user, HasManyAssociator.authorize_target(@association))
         associator = HasManyAssociator.new(@resource, @association, params)
         associator.perform
 
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -23,7 +23,8 @@ module ForestLiana
           format.json { render_jsonapi(getter) }
           format.csv { render_csv(getter, @association.klass) }
         end
-      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError => error
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
         # A CSV request already has its response Content-Type/Content-Disposition set by the
         # respond_to format match before this rescue ever runs — force JSON back, or the client
         # downloads a ".csv" file whose content is this JSON error.
@@ -61,6 +62,21 @@ module ForestLiana
         getter.count
 
         render serializer: nil, json: { count: getter.records_count }
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
+        render(serializer: nil, json: { errors: [{
+          status: error.error_code,
+          detail: error.message,
+          name: error.name,
+          data: error.data
+        }] }, status: error.status)
+      rescue ForestLiana::Errors::ExpectedError => error
+        error.display_error
+        error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
+          status: error.error_code,
+          detail: error.message
+        }])
+        render(serializer: nil, json: error_data, status: error.status)
       rescue => error
         FOREST_REPORTER.report error
         FOREST_LOGGER.error "Association Index Count error: #{error}\n#{format_stacktrace(error)}"
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -16,9 +16,7 @@ module ForestLiana
 
     def index
       begin
-        # Unlike ResourcesController#index, nothing upstream gates this route on any permission —
-        # matches agent-nodejs's list-related route, which checks browse (or export for CSV) on
-        # the foreign collection, not the parent, before listing it.
+        # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
         action = request.format == 'csv' ? 'export' : 'browse'
         forest_authorize!(action, forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
@@ -92,6 +90,10 @@ module ForestLiana
 
     def update
       begin
+        # BelongsToUpdater's writer saves the FK on the target for a has_one, on @resource only
+        # for a belongsTo.
+        edit_subject = @association.macro == :has_one ? @association.klass : @resource
+        forest_authorize!('edit', forest_user, edit_subject)
         updater = BelongsToUpdater.new(@resource, @association, params)
         updater.perform
 
@@ -101,6 +103,13 @@ module ForestLiana
         else
           head :no_content
         end
+      rescue ForestLiana::Errors::ExpectedError => error
+        error.display_error
+        error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
+          status: error.error_code,
+          detail: error.message
+        }])
+        render(serializer: nil, json: error_data, status: error.status)
       rescue => error
         FOREST_REPORTER.report error
         FOREST_LOGGER.error "Association Update error: #{error}\n#{format_stacktrace(error)}"
@@ -110,10 +119,18 @@ module ForestLiana
 
     def associate
       begin
+        forest_authorize!('edit', forest_user, @association.klass)
         associator = HasManyAssociator.new(@resource, @association, params)
         associator.perform
 
         head :no_content
+      rescue ForestLiana::Errors::ExpectedError => error
+        error.display_error
+        error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
+          status: error.error_code,
+          detail: error.message
+        }])
+        render(serializer: nil, json: error_data, status: error.status)
       rescue => error
         FOREST_REPORTER.report error
         FOREST_LOGGER.error "Association Associate error: #{error}\n#{format_stacktrace(error)}"
@@ -123,6 +140,8 @@ module ForestLiana
 
     def dissociate
       begin
+        action = (params[:delete].to_s == 'true' || HasManyDissociator.destroys_on_unlink?(@association)) ? 'delete' : 'edit'
+        forest_authorize!(action, forest_user, @association.klass)
         dissociator = HasManyDissociator.new(@resource, @association, params, forest_user)
         dissociator.perform
 
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -89,12 +89,13 @@ module ForestLiana
     end
 
     def update
+      # BelongsToUpdater's writer saves the FK on the target for a has_one, on @resource only
+      # for a belongsTo. Authorized outside the begin, like ResourcesController's own actions, so
+      # a denial reaches ApplicationController's rescue_from and keeps its name/data.
+      edit_subject = @association.macro == :has_one ? @association.klass : @resource
+      forest_authorize!('edit', forest_user, edit_subject)
+      forest_authorize!('delete', forest_user, @association.klass) if BelongsToUpdater.replaces_destructively?(@association)
       begin
-        # BelongsToUpdater's writer saves the FK on the target for a has_one, on @resource only
-        # for a belongsTo.
-        edit_subject = @association.macro == :has_one ? @association.klass : @resource
-        forest_authorize!('edit', forest_user, edit_subject)
-        forest_authorize!('delete', forest_user, @association.klass) if BelongsToUpdater.replaces_destructively?(@association)
         updater = BelongsToUpdater.new(@resource, @association, params)
         updater.perform
 
@@ -119,8 +120,8 @@ module ForestLiana
     end
 
     def associate
+      forest_authorize!('edit', forest_user, HasManyAssociator.authorize_target(@association))
       begin
-        forest_authorize!('edit', forest_user, HasManyAssociator.authorize_target(@association))
         associator = HasManyAssociator.new(@resource, @association, params)
         associator.perform
 
@@ -140,16 +141,16 @@ module ForestLiana
     end
 
     def dissociate
+      if params[:delete].to_s == 'true'
+        # Explicit delete destroys the far record directly, regardless of association shape.
+        action = 'delete'
+        authorize_target = @association.klass
+      else
+        action = HasManyDissociator.destroys_on_unlink?(@association) ? 'delete' : 'edit'
+        authorize_target = HasManyDissociator.destroy_target(@association)
+      end
+      forest_authorize!(action, forest_user, authorize_target)
       begin
-        if params[:delete].to_s == 'true'
-          # Explicit delete destroys the far record directly, regardless of association shape.
-          action = 'delete'
-          authorize_target = @association.klass
-        else
-          action = HasManyDissociator.destroys_on_unlink?(@association) ? 'delete' : 'edit'
-          authorize_target = HasManyDissociator.destroy_target(@association)
-        end
-        forest_authorize!(action, forest_user, authorize_target)
         dissociator = HasManyDissociator.new(@resource, @association, params, forest_user)
         dissociator.perform
 
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -16,6 +16,11 @@ module ForestLiana
 
     def index
       begin
+        # Unlike ResourcesController#index, nothing upstream gates this route on any permission —
+        # matches agent-nodejs's list-related route, which checks browse (or export for CSV) on
+        # the foreign collection, not the parent, before listing it.
+        action = request.format == 'csv' ? 'export' : 'browse'
+        forest_authorize!(action, forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.perform
 
@@ -58,6 +63,7 @@ module ForestLiana
       #         a double-render / 500 rather than the intended 404.
       return if performed?
       begin
+        forest_authorize!('browse', forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.count
 
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -140,8 +140,15 @@ module ForestLiana
 
     def dissociate
       begin
-        action = (params[:delete].to_s == 'true' || HasManyDissociator.destroys_on_unlink?(@association)) ? 'delete' : 'edit'
-        forest_authorize!(action, forest_user, @association.klass)
+        if params[:delete].to_s == 'true'
+          # Explicit delete destroys the far record directly, regardless of association shape.
+          action = 'delete'
+          authorize_target = @association.klass
+        else
+          action = HasManyDissociator.destroys_on_unlink?(@association) ? 'delete' : 'edit'
+          authorize_target = HasManyDissociator.destroy_target(@association)
+        end
+        forest_authorize!(action, forest_user, authorize_target)
         dissociator = HasManyDissociator.new(@resource, @association, params, forest_user)
         dissociator.perform
 
--- a/app/controllers/forest_liana/resources_controller.rb
+++ b/app/controllers/forest_liana/resources_controller.rb
@@ -31,7 +31,8 @@ module ForestLiana
       rescue ForestLiana::Errors::LiveQueryError => error
         render json: { errors: [{ status: 422, detail: error.message }] },
           status: :unprocessable_entity, serializer: nil
-      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError => error
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
         # A CSV request already has its response Content-Type/Content-Disposition set by the
         # respond_to format match before this rescue ever runs — force JSON back, or the client
         # downloads a ".csv" file whose content is this JSON error.
--- a/app/controllers/forest_liana/resources_controller.rb
+++ b/app/controllers/forest_liana/resources_controller.rb
@@ -182,6 +182,13 @@ module ForestLiana
         end
       rescue ActiveRecord::RecordNotDestroyed => error
         render json: { errors: [{ status: :bad_request, detail: error.message }] }, status: :bad_request
+      rescue ForestLiana::Errors::ExpectedError => error
+        error.display_error
+        error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
+          status: error.error_code,
+          detail: error.message
+        }])
+        render(serializer: nil, json: error_data, status: error.status)
       rescue => error
         FOREST_REPORTER.report error
         FOREST_LOGGER.error "Records Destroy error: #{error}\n#{format_stacktrace(error)}"
--- a/app/controllers/forest_liana/resources_controller.rb
+++ b/app/controllers/forest_liana/resources_controller.rb
@@ -69,6 +69,14 @@ module ForestLiana
       rescue ForestLiana::Errors::LiveQueryError => error
         render json: { errors: [{ status: 422, detail: error.message }] },
           status: :unprocessable_entity, serializer: nil
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
+        render(serializer: nil, json: { errors: [{
+          status: error.error_code,
+          detail: error.message,
+          name: error.name,
+          data: error.data
+        }] }, status: error.status)
       rescue ForestLiana::Errors::ExpectedError => error
         error.display_error
         error_data = ForestAdmin::JSONAPI::Serializer.serialize_errors([{
--- a/app/services/forest_liana/ability/permission.rb
+++ b/app/services/forest_liana/ability/permission.rb
@@ -308,12 +308,14 @@ module ForestLiana
       # guard raises its own, already-pinned message for it, and the query never runs either way.
       # Sort has no such downstream validator, so its own path is resolved without this rescue.
       #
-      # FiltersParser only ever reads one association hop (+assoc:field+); a deeper path such as
-      # +island:location:name+ still resolves here (FieldPath recurses as far as the reflections
-      # go), but the parser would reject it as a malformed/unknown field regardless of permission,
-      # so checking a collection the query itself can never reach is left unchecked too.
+      # FiltersParser#parse_field_name only ever reads one association hop (+assoc:field+) — it
+      # builds the quoted field from segment 1 but validates existence against the *last* segment,
+      # so a deeper path such as +island:name:id+ (a real column at 1, a real column at -1) slips
+      # its own validation and runs as a live filter on segment 1's table. Truncating to the same
+      # two segments the parser actually reads — mirroring +sort_field_path+ below — checks
+      # permission on the collection the query really touches, instead of skipping it outright.
       def query_usage(action, root_model, path)
-        return nil if path.count(':') > 1
+        path = path.split(':').first(2).join(':')
 
         { action: action, path: path, collections: resolve_owner(root_model, path) }
       rescue ForestLiana::Errors::HTTP422Error
--- a/app/services/forest_liana/ability/permission.rb
+++ b/app/services/forest_liana/ability/permission.rb
@@ -139,6 +139,28 @@ module ForestLiana
         redacted
       end
 
+      # Refused rather than redacted, unlike +redact_fields+: dropping a filter condition widens
+      # the result set, and dropping a sort clause silently reorders it. +root_model+ is pinned
+      # readable — +browse+/+read+ already gate it upstream — so it is never itself a refusal.
+      def assert_can_read_query_fields(user, root_model, filter_paths: [], sort_paths: [])
+        root_name = ForestLiana.name_for(root_model)
+
+        usages = filter_paths.filter_map { |path| query_usage('filter on', root_model, path) } +
+                 sort_paths.map { |path| { action: 'sort on', path: path, collections: resolve_owner(root_model, path) } }
+
+        return if usages.empty?
+
+        allowed = read_permissions(user, usages.flat_map { |usage| usage[:collections] }).merge(root_name => true)
+        readable_collection_names = allowed.filter_map { |name, ok| name if ok }
+
+        denied = usages.find { |usage| !FieldPath.readable_leaves?(usage[:collections], readable_collection_names) }
+        return unless denied
+
+        raise ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError.new(
+          denied[:action], denied[:path], denied[:collections]
+        )
+      end
+
       def is_chart_authorized?(user, parameters)
         parameters = parameters.to_h
         parameters.delete('timezone')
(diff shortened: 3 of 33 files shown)


app/services/forest_liana/filters_parser.rb
```diff
@@ -2,6 +2,15 @@ module ForestLiana
   class FiltersParser
     AGGREGATOR_OPERATOR = %w(and or).freeze
 
+    # The `field` of every leaf in a filter tree, however deeply nested — the same tree
+    # `apply_filters` will walk, read rather than re-derived.
+    def self.field_paths(filter)
+      return [] if filter.nil?
+      return filter['conditions'].flat_map { |condition| field_paths(condition) } if filter['aggregator']
+
+      [filter['field']]
```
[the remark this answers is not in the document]
matthv: Fixed — `field_paths` now only forwards a leaf whose `field` is a String; anything else (missing, or a non-String value) is left out so FiltersParser's own `ensure_valid_condition` rejects it with its usual 422 instead of this guard crashing on it first. Added specs, including an HTTP-level one reproducing the exact reported case.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -20,7 +22,9 @@ def perform(resource)
         ForestLiana::QueryHelper.get_tables_associated_to_relations_name(@resource)
       @records = search_param
 
-      filters = ForestLiana::ScopeManager.append_scope_for_user(@params[:filters], @user, @collection.name)
+      caller_filter = @params[:filters].present? ? ForestLiana::ScopeManager.inject_context_variables(@params[:filters], @user) : nil
```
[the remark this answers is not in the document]
matthv: False positive, traced empirically. `@params[:filters]` is the raw string `"{}"`, which is `.present?` — the ternary never produces `nil`. And `ScopeManager.inject_context_variables` parses it then runs `ContextVariablesInjector.inject_context_in_filter`, which turns `{}` into `{'field'=>nil,'operator'=>nil,'value'=>nil}` (not empty anymore) before it ever reaches `append_scope`/`blank?`. That still hits `ensure_valid_condition`'s type check and raises the same 422 `Invalid condition format` as before this PR — confirmed with a live request, `STATUS: 422`.

app/controllers/forest_liana/resources_controller.rb
```diff
@@ -27,7 +27,8 @@ def index
       rescue ForestLiana::Errors::LiveQueryError => error
         render json: { errors: [{ status: 422, detail: error.message }] },
           status: :unprocessable_entity, serializer: nil
-      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError => error
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
```
christophebrun-forest, line 35: **Blocking — `destroy_bulk` turns this 403 into a bodiless 500.**

Verified against a running app: a select-all delete carrying a filter on an unreadable collection returns **`500` with an empty body**, plus a `FOREST_REPORTER.report` (so it pages someone) — because `destroy_bulk` (line 170) has a blanket `rescue => error` with no `ExpectedError` clause ahead of it.

```
PROBE status=500 body=""
```

The description says `destroy_bulk` going through the guard "with no exemption" is deliberate, and I agree — the denial is right. It's the *response* that isn't, and it's wrong in exactly the way this PR calls out as the bug it fixes elsewhere ("had no `ExpectedError` rescue at all before this — any denial surfaced as a bodiless 500").

Same clause needed before `destroy_bulk`'s blanket rescue.

christophebrun-forest, line 35: **`resources#count` renders the 403 without `name`/`data`.**

The description claims the error is wired into "the four routes… index, count, and their relationships equivalents", but this file only got the clause on `index`. `count` falls through to the generic `ExpectedError` rescue (line 64):

```
count : {"errors":[{"status":403,"detail":"You cannot filter on 'island:name': …"}]}
index : {"errors":[{"status":403,"detail":"…","name":"UnauthorizedQueryFieldError","data":{"action":"filter on","field":"island:name"}}]}
```

So three of the four routes carry `name`/`data` and one doesn't — and that route also emits an `error.display_error` log line for a legitimate 403.

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -53,6 +54,21 @@ def count
         getter.count
 
         render serializer: nil, json: { count: getter.records_count }
+      rescue ForestLiana::Ability::Exceptions::UnauthorizedFieldsError,
+             ForestLiana::Ability::Exceptions::UnauthorizedQueryFieldError => error
```
christophebrun-forest, line 74: Good fix here (and the `ExpectedError` rescue added just below also repairs, as a side effect, the `FiltersParser` 422s that used to surface as 500s on this route).

**But `dissociate` (line 112) has the same shape as the one you're fixing here**: `HasManyDissociator` → `get_ids_from_request` → the guard, then a blanket `rescue => error` → bodiless 500 + report. Worth handling alongside `destroy_bulk`.

app/services/forest_liana/ability/permission.rb
```diff
@@ -133,6 +133,28 @@ def redact_fields(user, root_model, fields_hash, named_collections:)
...
+      # readable — +browse+/+read+ already gate it upstream — so it is never itself a refusal.
+      def assert_can_read_query_fields(user, root_model, filter_paths: [], sort_paths: [])
+        root_name = ForestLiana.name_for(root_model)
+
+        usages = filter_paths.filter_map { |path| query_usage('filter on', root_model, path) } +
+                 sort_paths.map { |path| { action: 'sort on', path: path, collections: resolve_owner(root_model, path) } }
+
+        return if usages.empty?
+
+        allowed = read_permissions(user, usages.flat_map { |usage| usage[:collections] }).merge(root_name => true)
```
christophebrun-forest: **Blocking — one extra network round trip per request for a `browse`-without-`read` role.**

`root_name` is handed to `read_permissions`, which doesn't know it's pinned: it sees it denied and triggers the retry-on-denial refetch — `get_collections_permissions_data(true)` deletes the process-wide `forest.collections` cache and re-hits Forest.

Counting `/liana/v4/permissions/environment` fetches for a role with `browse` but not `read` on the root, listing with a filter and a sort on its own columns:

```
with the PR:          request#1 → 2   #2 → 3 (cumulative 5)   #3 → 7    (+2/request)
baseline (no filter/sort, guard short-circuits):                        (+1/request)
excluding root_name:  request#1 → 2   #2 → 3                  #3 → 4    (+1/request)
```

This is the exact config you added a spec for ("a role can legitimately browse a collection without having its own `read`"), so it isn't hypothetical. And it's a successful 200 that invalidates a shared cache and makes a network call every time — worse than a denial, which at least stops.

One line, since the root is pinned readable right after anyway:

```ruby
allowed = read_permissions(user, usages.flat_map { |usage| usage[:collections] }.uniq - [root_name]).merge(root_name => true)
```

Applied it and re-ran the full suite: 596 examples, 0 failures.

christophebrun-forest (commented):
## Review

I checked out the branch, set up the dummy DB and ran the suite: **596 examples, 0 failures**. I then probed the wiring with throwaway request specs — everything below is verified against a running app, not read off the diff.

### What's right

The core design holds up. Three things I specifically tried to break and couldn't:

- **Refuse-not-redact** is the correct call, and the reasoning in the description is sound.
- **The "read the caller's tree once" seam.** Splitting `append_scope_for_user` into `append_scope` so the guard authorizes *the same object* `FiltersParser` applies is the right move — a second parse could drift, and re-injecting `{{...}}` placeholders would raise on a value that merely looks like one. The "byte for byte" spec pins it.
- **Sort skipped on count.** I confirmed Rails' `execute_simple_calculation` does `unscope(:order)` when there's no GROUP BY, so the ORDER BY genuinely never reaches the COUNT. And the unchecked `get_ids_from_request` path is harmless for a different reason worth knowing: `fetch_ids` uses `find_in_batches`, which forces batch order and discards the scoped order entirely — so a sort on a secret column can't shift *which* ids a capped select-all returns.

Specs are thorough and the mutation-checking claim is credible.

### Verdict

Two blocking points to my eye, detailed inline:

1. **`destroy_bulk` and `dissociate` turn the new 403 into a bodiless 500** — exactly the defect this PR fixes for `associations#count`, with a `FOREST_REPORTER.report` on top.
2. **One extra permissions-API round trip per list request** for a `browse`-without-`read` role — the exact config this PR added a spec for. One-line fix, verified suite green.

Then two small ones worth folding in: `resources#count` renders the 403 without `name`/`data` (the description does claim all four routes), and `FiltersParser.field_paths` crashes into a 500 on a malformed aggregator node where the parser used to answer 422.

*(Posted as `COMMENT` rather than `REQUEST_CHANGES` so as not to block the stack — happy to escalate.)*

spec/requests/resources_spec.rb
```diff
@@ -247,6 +248,107 @@
...
+      get '/forest/Tree', params: params, headers: headers
+
+      expect(response.status).to eq(200)
+    end
+
+    it 'never refuses a filter on the root collection, even when the root has no read permission of its own' do
+      # `browse` (not `read`) is what forest_authorize! gates the route on; a role can legitimately
+      # browse a collection without having its own `read` — the root is still pinned readable for
+      # this guard, which must not re-derive a denial for it from a permission it is not gated on.
+      Rails.cache.write('forest.collections', { 'Tree' => { 'browse' => [1], 'read' => [], 'edit' => [], 'add' => [], 'delete' => [], 'export' => [], 'actions' => {} } })
```
christophebrun-forest: This spec currently passes for the wrong reason. It writes `read => []` here, but the forced refetch that this very denial triggers throws that value away and re-reads the outer `before` stub, which *grants* `Tree` read. The example would stay green with the root pin removed — it isn't testing the pin.

Stubbing the source with root read denied would make it bite (and would make the refetch described in my `permission.rb` comment visible).

spec/requests/resources_spec.rb
```diff
@@ -247,6 +248,107 @@
...
+        get '/forest/Tree', params: params, headers: headers
+
+        expect(response.status).to eq(403)
+        body = JSON.parse(response.body)
+        expect(body['errors'][0]['detail'])
+          .to eq "You cannot filter on 'island:name': you are not allowed to read the 'Island' collection."
+        expect(body['errors'][0]['data']).to eq('action' => 'filter on', 'field' => 'island:name')
+      end
+
+      it 'refuses count the same way' do
```
christophebrun-forest: This assertion is what lets the shape divergence between `index` and `count` through: the sibling `index` example asserts the body, this one only the status. Asserting `name`/`data` here too would turn the missing rescue clause on `resources#count` red.

app/services/forest_liana/filters_parser.rb
```diff
@@ -2,6 +2,17 @@ module ForestLiana
   class FiltersParser
     AGGREGATOR_OPERATOR = %w(and or).freeze
 
+    # The `field` of every leaf in a filter tree, however deeply nested — the same tree
+    # `apply_filters` will walk, read rather than re-derived. A leaf with no valid `field` is left
+    # out rather than forwarded: `ensure_valid_condition` raises its own 422 for it right after
+    # this is read, and a non-String value would otherwise reach FieldPath, which expects one.
+    def self.field_paths(filter)
+      return [] if filter.nil?
+      return filter['conditions'].flat_map { |condition| field_paths(condition) } if filter['aggregator']
```
christophebrun-forest: Crashes on a malformed aggregator node:

```ruby
FiltersParser.field_paths({ 'aggregator' => 'and', 'conditions' => { 'field' => 'name' } })
# => TypeError: no implicit conversion of String into Integer  → 500
```

Before this PR, `parse_aggregation` walked the same shape and produced a clean `422 'Filters cannot be a raw value'`. This is the mirror image of the case 6eaec17 already fixed for `field`, so it fits the intent you already committed to:

```ruby
return [] unless filter['conditions'].is_a?(Array)
```

(`{'aggregator' => 'and'}` with no `conditions` at all was already a 500 on `main`, so that one isn't a regression.)

app/controllers/forest_liana/resources_controller.rb
matthv: Fixed in d9e952a — same `ExpectedError` rescue clause added before `destroy_bulk`'s blanket rescue. Test added (403 + body, not the bodiless 500).

matthv: Fixed in 40a0dcb — same UnauthorizedFieldsError/UnauthorizedQueryFieldError rescue clause added to count, before its ExpectedError fallback. Strengthened the sibling spec too (asserts name/data now, not just status).

app/controllers/forest_liana/associations_controller.rb
matthv: Fixed in d9e952a, same clause as destroy_bulk. Test added on the select-all dissociate path.

app/services/forest_liana/ability/permission.rb
matthv: Fixed in ccf8647 — root_name excluded from what's handed to read_permissions, then merged back in as always-true. Test counts real get_permissions calls for this exact role config (browse without read on root) and mutation-checked (2 calls without the fix, 0 with it).

app/services/forest_liana/filters_parser.rb
matthv: Fixed in 84263f0 — but that alone wasn't enough: the same 'conditions assumed to be an Array' bug turned out to exist at two more sites on this exact request path (ContextVariablesInjector#inject_context_in_filter, ResourcesGetter#extract_associations_from_filter), both crashing before field_paths is even reached. Fixed all three, then in b90e145 pushed the check into FiltersParser#ensure_valid_aggregation itself, since the three call-site guards only degraded gracefully for a Hash-shaped conditions — a nil one still crashed uncaught. Added a nil-conditions case and a nested-malformed-aggregation case.

spec/requests/resources_spec.rb
matthv: Fixed in 40a0dcb, same commit as the count name/data fix above.

spec/requests/resources_spec.rb
matthv: Not fixed — left this one as-is, out of scope for this round. FWIW, checked whether the assertion itself is still meaningful post-fix: reverting the .merge(root_name => true) override does turn this test red (403), so it does catch a removal of the pin. But your point about the setup stands independently — it still doesn't exercise the source-stubbed persistent-denial path the way the sibling tests in this file now do. Happy to rewrite it properly if you'd rather not merge with it as-is.

app/services/forest_liana/ability/permission.rb
```diff
@@ -133,6 +133,31 @@ def redact_fields(user, root_model, fields_hash, named_collections:)
...
+
+        usages = filter_paths.filter_map { |path| query_usage('filter on', root_model, path) } +
+                 sort_paths.map { |path| { action: 'sort on', path: path, collections: resolve_owner(root_model, path) } }
+
+        return if usages.empty?
+
+        # root_name is pinned readable above already; leaving it in would make a denial for it
+        # (the only way it could ever appear in usages: an owner resolves back to the root itself)
+        # trigger read_permissions' retry-on-denial refetch on every single request.
+        allowed = read_permissions(user, usages.flat_map { |usage| usage[:collections] }.uniq - [root_name]).merge(root_name => true)
```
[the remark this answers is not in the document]
matthv: Confirmed and fixed in 6f844bc — reproduced the leak first (a role with zero permission on the related collection got a real filtered count back, a binary-searchable filter oracle). Added forest_authorize! on the related collection to both index and count, matching agent-nodejs's list-related/count-related routes (assertCanBrowse on the foreign collection, not the parent).

spec/requests/resources_spec.rb
matthv: Fixed in 6c08105 — stubbed at the source (persistently denied) instead of writing to the derived cache, same pattern as every other test in this file. Mutation-checked: removing the root pin now correctly turns this test red.

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -93,12 +132,21 @@ def associate
 
     def dissociate
       begin
+        action = (params[:delete].to_s == 'true' || HasManyDissociator.destroys_on_unlink?(@association)) ? 'delete' : 'edit'
```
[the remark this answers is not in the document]
matthv: Fixed in b6c394c — but the fix is broader than the suggested patch. You're right that dependent: :destroy on a through association only destroys join rows, never the far collection — but a plain unlink with NO dependent option at all *also* destroys the join row (Rails' HasManyThroughAssociation#delete_records falls through to scope.delete_all by default), and that case was checking nothing at all before this fix, not just the wrong collection. destroys_on_unlink?/destroy_target now check the join reflection's collection whenever anything gets unlinked (dependent: :nullify is the only through case that doesn't destroy it), verified against a real has_many :through fixture rather than the far collection. Confirmed real has_and_belongs_to_many isn't caught by this branch either (macro: :has_and_belongs_to_many, options[:through] nil).

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -80,10 +111,18 @@ def update
 
     def associate
       begin
+        forest_authorize!('edit', forest_user, @association.klass)
```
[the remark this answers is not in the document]
matthv: Fixed in f57435c, same fix shape as the dissociate one above (HasManyAssociator.authorize_target instead of a controller-level ternary).

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -62,6 +82,10 @@ def count
 
     def update
       begin
+        # BelongsToUpdater's writer saves the FK on the target for a has_one, on @resource only
+        # for a belongsTo.
+        edit_subject = @association.macro == :has_one ? @association.klass : @resource
+        forest_authorize!('edit', forest_user, edit_subject)
```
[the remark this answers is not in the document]
matthv: Fixed, but gated rather than unconditional: replacing a has_one target only destroys the previous one when the reflection's own dependent option is :destroy or :delete — confirmed empirically that Rails just nullifies the FK otherwise, leaving the old record alive. Requiring delete on every has_one update (as suggested) would have over-restricted that default case. BelongsToUpdater.replaces_destructively? gates it on the actual option instead, tested against a real has_one with dependent: :destroy.

app/services/forest_liana/belongs_to_updater.rb
```diff
@@ -4,6 +4,12 @@ class BelongsToUpdater
 
     attr_accessor :errors
 
+    # Replacing a has_one target only destroys the previous one when the reflection's own
+    # dependent option says so (:destroy or :delete) — the default just nullifies its FK.
+    def self.replaces_destructively?(association)
+      association.macro == :has_one && %i[destroy delete].include?(association.options[:dependent])
```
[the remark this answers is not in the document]
matthv: Fixed in af63483 — excluded has_one :through from the destructive-replace check. Tried to build a live, assignable has_one :through fixture to test it end-to-end but hit Rails' own structural constraints twice (through a has_one, through a has_many source) before it would even accept the shape; covered with a double-based unit test instead.

app/services/forest_liana/has_many_dissociator.rb
```diff
@@ -2,6 +2,21 @@ module ForestLiana
...
+    def self.destroys_on_unlink?(association)
+      return association.options[:dependent] != :nullify if association.options[:through]
+
+      association.macro == :has_many && %i[destroy delete_all].include?(association.options[:dependent])
+    end
+
+    # The collection a plain unlink actually writes to: the join collection for a through
+    # association (see destroys_on_unlink?), the far collection otherwise.
+    def self.destroy_target(association)
+      association.options[:through] ? association.through_reflection.klass : association.klass
```
christophebrun-forest: **A `has_many :through` whose join model is hidden from the schema now 409s — the feature stops working rather than getting gated.**

Targeting the join collection is right in principle, and the Rails reasoning behind `destroys_on_unlink?` checks out (I read `HasManyThroughAssociation#delete_records`: with no `dependent` it falls through to `scope.delete_all` on the join rows, so `!= :nullify` is correct, `dependent: nil` included). But it assumes the join model *is* an authorizable Forest collection. `ForestLiana.excluded_models = ['Membership']` is the normal way to hide a join table from the UI, and then:

```
dissociate → 409 {"errors":[{"status":409,"detail":"The collection Membership doesn't exist"}]}
             env_fetches=2   join_row_still_there=true
associate  → 409 (same)
```

`is_crud_authorized?` misses the key, force-refetches the whole permissions cache, then dereferences `nil[action]` → `UnknownCollection` (409). Both routes worked before these commits — unauthorized, which was the hole — but for those deployments this doesn't *gate* the feature, it removes it, and pays a permissions refetch per attempt on the way out.

A fallback when the join isn't exposed keeps the pre-existing behaviour there while preserving the new precision when it is:

```ruby
def self.destroy_target(association)
  join = association.options[:through] && association.through_reflection.klass
  join && SchemaUtils.model_included?(join) ? join : association.klass
end
```

Same shape needed on `HasManyAssociator.authorize_target` (line 8), which has the identical ternary. No spec covers this case today.

For context, everything else in this round checks out on my side: all four earlier findings are fixed (`b90e145` in particular fixes more than I suggested — my `field_paths` one-liner would have left `parse_aggregation` crashing on `conditions: nil`), and the suite is green at 638 examples / 0 failures. One cosmetic leftover: `forest_authorize!` sits *inside* the `begin` block on these routes, so an `AccessDenied` is rendered by the local `ExpectedError` branch instead of `ApplicationController`'s `rescue_from`, and loses `name: "AccessDenied"` that `resources_controller` still returns (which calls it outside the block).

Worth a release note too: `associations#index`/`#count` now requiring `browse` on the related collection is the correct target state, but it 403s roles that could previously see a related list without that permission.

matthv: Fixed in 27a2372 — destroy_target/authorize_target now check SchemaUtils.model_included?(join) first, falling back to the far collection (the pre-existing check) when the join model isn't an authorizable collection. Reproduced your exact scenario with excluded_models = ['Membership'] and mutation-checked (409 without the fallback, correct 403/204 with it). Same fix on both HasManyDissociator and HasManyAssociator as you flagged.

Also picked up the two other things you noted: forest_authorize! now runs before the begin on update/associate/dissociate too (9976457), so a denial keeps name: "AccessDenied" like index/count already do — and added a behavioral-change note to the PR description about associations#index/#count now requiring browse on the related collection.

christophebrun-forest (commented):
**`9976457` fixed 3 of the 5 routes — `index` and `count` still hold `forest_authorize!` inside their `begin`.**

Its commit message says index/count "already return" `name: "AccessDenied"`, but they don't: they got their `forest_authorize!` in the same batch (`6f844bc`), in the same place, so they have the same defect. On the branch as it stands:

```
assoc index      403 {"status":403,"detail":"You don't have permission…"}
assoc count      403 {"status":403,"detail":"You don't have permission…"}
assoc update     403 {"status":403,"detail":"…","name":"AccessDenied"}
assoc associate  403 {"status":403,"detail":"…","name":"AccessDenied"}
assoc dissociate 403 {"status":403,"detail":"…","name":"AccessDenied"}
resources index  403 {"status":403,"detail":"…","name":"AccessDenied"}
```

Four suggestions below: the same move you already made, on the two remaining routes, plus the assertions whose absence let this slip past both rounds.

Point A is closed on my side — `27a2372` is exactly the fallback, and the `around` block that really sets `ForestLiana.excluded_models` plus the 403/204 pair covers both outcomes. Suite green at 643 examples / 0 failures.

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -12,24 +12,24 @@ class AssociationsController < ForestLiana::ApplicationController
 
     def index
       begin
+        # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
+        action = request.format == 'csv' ? 'export' : 'browse'
+        forest_authorize!(action, forest_user, @association.klass)
```
christophebrun-forest: Safe to hoist here: `find_resource`/`find_association` are `before_action`s, and a `render` inside a `before_action` **halts** the chain — so `@association` can never be `nil` by the time `index` runs. That's exactly why `count`, which calls them by hand, needed its own `return if performed?` guard and `index` never did.

```suggestion
      # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
      # Authorized outside the begin, like ResourcesController's own actions, so a denial reaches
      # ApplicationController's rescue_from and keeps its name/data.
      action = request.format == 'csv' ? 'export' : 'browse'
      forest_authorize!(action, forest_user, @association.klass)
      begin
```

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -53,10 +53,26 @@ def count
       #         a double-render / 500 rather than the intended 404.
       return if performed?
       begin
+        forest_authorize!('browse', forest_user, @association.klass)
```
christophebrun-forest: Here the placement matters: it has to stay **after** `return if performed?`, never before it. Hoisted above that guard, an association name that doesn't resolve would dereference `nil.klass` and turn the intended 404 back into the 500 the guard was added to prevent.

```suggestion
      forest_authorize!('browse', forest_user, @association.klass)
      begin
```

spec/requests/associations_spec.rb
```diff
@@ -81,33 +82,462 @@
...
+        )
+      Rails.cache.delete('forest.collections')
+    end
+
+    it 'refuses index with a 403, rather than serving it unauthorized' do
+      get "/forest/Island/#{@island.id}/relationships/trees",
+          params: { page: { 'number' => '1', 'size' => '10' }, timezone: 'Europe/Paris' }, headers: headers
+
+      expect(response.status).to eq(403)
+      expect(JSON.parse(response.body)['errors'][0]['detail']).to eq "You don't have permission to access this resource"
```
christophebrun-forest, line 182: This is the assertion gap that let the shape divergence through twice — status and detail are both already correct on this route, `name` is the only thing missing.

```suggestion
      expect(response.status).to eq(403)
      body = JSON.parse(response.body)
      expect(body['errors'][0]['detail']).to eq "You don't have permission to access this resource"
      expect(body['errors'][0]['name']).to eq 'AccessDenied'
```

spec/requests/associations_spec.rb
```diff
@@ -81,33 +82,462 @@
...
+    end
+
+    # Without forest_authorize!, a filtered count on a collection the role cannot even browse
+    # answered with the real, filtered row count — letting the filter value be guessed one probe
+    # at a time (0 vs a positive count) without ever touching a field-level read check.
+    it 'refuses count with a 403, instead of a filtered count that leaks whether a value matches' do
+      params = { filters: JSON.generate({ 'field' => 'name', 'operator' => 'equal', 'value' => 'Lemon Tree' }), timezone: 'Europe/Paris' }
+      get "/forest/Island/#{@island.id}/relationships/trees/count", params: params, headers: headers
+
+      expect(response.status).to eq(403)
```
christophebrun-forest, line 193: Same here, and this one asserts only the status today:

```suggestion
      expect(response.status).to eq(403)
      body = JSON.parse(response.body)
      expect(body['errors'][0]['detail']).to eq "You don't have permission to access this resource"
      expect(body['errors'][0]['name']).to eq 'AccessDenied'
```

app/services/forest_liana/ability/permission.rb
```diff
@@ -139,6 +139,31 @@ def redact_fields(user, root_model, fields_hash, named_collections:)
         redacted
       end
 
+      # Refused rather than redacted, unlike +redact_fields+: dropping a filter condition widens
+      # the result set, and dropping a sort clause silently reorders it. +root_model+ is pinned
+      # readable — +browse+/+read+ already gate it upstream — so it is never itself a refusal.
+      def assert_can_read_query_fields(user, root_model, filter_paths: [], sort_paths: [])
+        root_name = ForestLiana.name_for(root_model)
+
+        usages = filter_paths.filter_map { |path| query_usage('filter on', root_model, path) } +
```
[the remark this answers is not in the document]
matthv: Fixed in 8f9f901 — `query_usage` now stops at one association hop, matching `FiltersParser`'s own depth; a deeper path is left unchecked here for the parser's own validation error, same treatment as an unresolvable one just above.

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -12,14 +12,18 @@ class AssociationsController < ForestLiana::ApplicationController
 
     def index
       begin
+        # Parity with agent-nodejs's list-related route: browse/export on the foreign collection.
+        action = request.format == 'csv' ? 'export' : 'browse'
+        forest_authorize!(action, forest_user, @association.klass)
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.perform
```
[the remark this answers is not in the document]
matthv: Fixed in 85211b4 — same move as update/associate/dissociate, hoisted outside the begin for index and count too.

matthv:
Fixed in 85211b4 — index and count now authorize outside their begin too, same shape as update/associate/dissociate. Pinned with the same `name: 'AccessDenied'` assertion the other three already had.

app/services/forest_liana/has_many_dissociator.rb
matthv: Fixed in 27a2372 — falls back to the far collection (User) when the join model is excluded from the schema via ForestLiana.excluded_models, covered by the around block toggling that plus the 403/204 pair for associate and dissociate.

app/controllers/forest_liana/associations_controller.rb
matthv: Fixed in 85211b4, same placement.

app/controllers/forest_liana/associations_controller.rb
matthv: Fixed in 85211b4, kept after `return if performed?` as you flagged.

spec/requests/associations_spec.rb
matthv: Added in 85211b4.

spec/requests/associations_spec.rb
matthv: Added in 85211b4.

christophebrun-forest (approved):
OK for me for merging in feature branch.

Note : 
The description no longer matches the code after 2064960 — three claims to update:

- "resolves each filter/sort path to **its leaf collection**" — it resolves segment 1 only now, which is the whole point of that commit.
- "a sort's has no such downstream validator, so it is left to raise directly from `FieldPath` instead of silently letting a bogus sort through to a 500" — no longer true: `sort=nosuchrel.name` returns a bodiless 500. Not a regression (`main` does the same), but the sort path has no validator at all now, so this was a deliberate behaviour that quietly disappeared — worth either dropping the paragraph or restoring the check.
- "the filter-path-skip-vs-sort-path-raise asymmetry [is] mutation-checked" — that asymmetry no longer exists in either direction.

Example counts are stale too (588 → 654).

app/services/forest_liana/belongs_to_updater.rb
```diff
@@ -4,6 +4,14 @@ class BelongsToUpdater
 
     attr_accessor :errors
 
+    # Replacing a has_one target only destroys the previous one when the reflection's own
+    # dependent option says so (:destroy or :delete) — the default just nullifies its FK.
+    # Rails ignores dependent: on a has_one :through, so that option is never destructive there.
+    def self.replaces_destructively?(association)
+      association.macro == :has_one && !association.options[:through] &&
+        %i[destroy delete].include?(association.options[:dependent])
```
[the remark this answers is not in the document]
matthv: Verified against Rails 6.1's own `has_one_association.rb`: `replace` calls `remove_target!(options[:dependent])`, whose `case` only handles `:delete`/`:destroy` — no `:destroy_async` branch, so it falls to the `else` (nullify). The async-enqueue logic (`enqueue_destroy_association`) only lives in `delete`/`handle_dependency`, run when the *owner* is destroyed, not when the association is reassigned. Confirmed live: replacing a `has_one dependent: :destroy_async` target does not destroy it, only nullifies the FK — same as no `dependent:` at all. Adding `:destroy_async` here would incorrectly require `delete` for an operation that never destroys anything.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -20,7 +22,9 @@ def perform(resource)
...
+      assert_can_read_query_fields(@user, root_model, filter_paths: FiltersParser.field_paths(caller_filter))
```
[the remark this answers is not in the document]
matthv: Fixed in 8735ad3 — the actual crash site was a step earlier though: `ContextVariablesInjector.inject_context_in_filter` calls `filter.key?('aggregator')` before `FiltersParser.field_paths` is ever reached, so that's where the guard had to go (plus `field_paths` itself as defense-in-depth). A non-blank non-Hash filter still reaches the parser's 422 as intended; an empty-Array filter degrades to "no filter" (pre-existing `.blank?` behavior, unrelated to this PR).

PR #797, opened 2026-09-03
feat(permissions): check what a search reads, without refusing a search whose reach can't be described

matthv:
Third and last enforcement guard of the "read-permission parity with v2" project (after PRD-1083's redaction and PRD-1084's filter/sort refusal). Stacked on #796 (PRD-1084) — base branch is `feature/optim-rbac-capabilities`, not `main`.

## Why

A search term is compared against columns the caller never names, and nothing checked that set: a plain search leaks a column's values one guess per request, and an extended search reaches into relations the same way a filter does.

## What

`SearchQueryBuilder#search_param` now records the path of every condition it adds through a single `push_condition` call, so the footprint reported to the permission guard can never drift from the SQL actually generated — pinned by a spec that independently derives both sets and compares them. That footprint feeds the same `assert_can_read_query_fields` filter/sort already use (`search_paths:` kwarg, action `'search on'`), which also now distinguishes a collection **absent from the apimap** ("unexposed", no role can ever be granted read on it) from one merely **denied** — a message fix that benefits the filter/sort guards from PRD-1084 too.

A smart-field `:search` lambda's own reach is deliberately **not** checked, in either plain or extended search: it can read anything, but it runs identically regardless of the `searchExtended` param, so refusing it in one mode only would cost every customer of that hook their extended search without closing anything a caller couldn't already reach on the default path. Documented as an explicit boundary rather than gated behind a refusal.

## Behavior change to verify

A search term matching no real column (e.g. `search_fields` naming only a smart field with no `search:` lambda, or only dotted/relation entries) now answers **no records**, not the whole table as before. This is the correct side of the previous bug — a config nobody would call safe was silently serving everything — but it does change the response of any live agent whose `search_fields` has this shape, so flagging it here rather than leaving it to be discovered in a "What" paragraph.

## Review rounds

Several passes of human review (Christophe Brun) and Macroscope, each fixed and mutation-checked before the next:

- A failed smart-search lambda no longer leaves the search unfiltered (was: exception logged and swallowed, whole table returned), then — once a single flag conflated "a lambda ran" with "a lambda actually filtered" — a *no-op* lambda (one returning its query untouched) stopped correctly emptying a malformed-UUID search with no column match. Fixed by tracking the two contributors (`@conditions_pushed`, `@lambda_contributed`) separately.
- `search_fields` naming a to-many association the agent doesn't expose is no longer searched (was: searched, then refused with a message about the *caller's* permissions for a *developer's* config).
- `resources#destroy_bulk` and `associations#dissociate` (both reach `get_ids_from_request`, which can raise any of these four exceptions) now return 403 instead of a bodiless 500, via a shared `QUERY_PERMISSION_ERRORS` constant + `rescue *…; raise` replacing the render block duplicated across all six call sites.
- `acts_as_taggable_query`'s subquery is deferred (`.to_sql`) past the permission check instead of executed eagerly ahead of it, then table-qualified (an unqualified `SELECT id` was ambiguous once joined against `taggings`), then kept out of the same `where(sql, binds)` call as the rest of the conditions (a tag name containing a colon crashed the whole search otherwise — Rails' named-bind scanner has no notion of "this colon is inside an already-quoted SQL literal").
- A malformed-UUID search that also happened to match a real id/enum/tag was being emptied unconditionally at one point in this history; fixed to still serve that match, same as before any of this PR's changes.

## Known gaps (pre-existing, not introduced here)

- `acts_as_taggable_on` isn't installed in the dummy app; several of the fixes above are guaranteed by construction (documented in comments next to them) rather than spec-covered against a real taggable model — stubbing `taggable?` on the real `Tree` model permanently corrupts ActiveRecord's relation-delegation cache for the rest of the process, regardless of stubbing method. One side effect worth knowing: `push_condition` now always receives a non-nil string for a taggable resource, so `@conditions_pushed` is unconditionally true for one regardless of whether the term actually matched a tag (harmless — a non-matching subquery still yields zero rows — but distinct from every other branch's meaning of that flag). Tracked as PRD-1190 for real CI coverage.
- `destroy_bulk`/`dissociate`'s fix isn't covered by a request-level spec — both route through a custom `ForestLiana::Router` that made a from-scratch spec setup a bigger lift than the fix itself; the underlying `raise`/`rescue_from` mechanism is exercised by the (many) existing `index`/`count` specs.
- A polymorphic relation whose targets are exposed but denied, mixed with one target that's unexposed, isn't separately spec'd (unit tests cover each mechanic individually).

## Verification

774 examples, 0 failures, green across every supported Rails version (6.1 through 8.1). Every guard (footprint↔SQL pin, unexposed-vs-denied, match-none, the review-round fixes, the two-contributor split) has a mutation-check: reverting the relevant line turns the corresponding spec red.

fixes PRD-1085

> [!NOTE]
> ### Add search-path read checks and `UnexposedQueryCollectionError` to `Permission`
> - `Permission#assert_can_read_query_fields` now checks search paths for read permission, in addition to filter and sort paths. Denied targets absent from the apimap raise the new `UnexposedQueryCollectionError` instead of a generic unauthorized-field error.
> - `SearchQueryBuilder` records the field paths accessed by generated search SQL so the permission guard can check them. Tag searches now generate a lazy SQL subquery instead of materializing IDs in Ruby.
> - `ResourcesController` and `AssociationsController` re-raise query-permission exceptions to `ApplicationController` for standard error rendering instead of handling them locally.
> - Risk: `ResourcesGetter#compute_includes` no longer eager-loads has-many associations inferred only from dotted search fields in plain queries.
>
> 
> #### Changes since #797 opened
>
> - Updated test teardown blocks to clear dynamically-added ActiveRecord polymorphic associations using both string and symbol keys, and explicitly invoke the reflections cache clearing method [7f9c38f]
> - Added detection and reporting of unexposed collections in permission checks [d9e6a80]
> - Changed Smart Action scope validation to re-raise expected errors [d9e6a80]
> - Added test coverage for unexposed collection handling [d9e6a80]
> - Updated inline documentation in search and permission code [d9e6a80]
> - Expanded documentation in `ForestLiana::SearchQueryBuilder` [1643a00]
> - Modified permission denial tracking to distinguish between unexposed and merely-denied collections [30e1b39]
> - Updated error data structure and message formatting for unauthorized field access [30e1b39]
> - Added test coverage for new permission denial scenarios [30e1b39]
> - Updated in-line commentary in `ForestLiana.SearchQueryBuilder` class to clarify that every case either constrains or empties the result set [60c6275]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c9e4f54.</sup>
> 
>


app/services/forest_liana/search_query_builder.rb
```diff
@@ -74,37 +90,43 @@ def search_param
...
+        # covered by a live spec — acts_as_taggable_on isn't installed in the dummy app, and
+        # stubbing `taggable?`/`acts_as_taggable` on the real Tree model to simulate it permanently
+        # corrupts ActiveRecord::Delegation's per-class method cache for the rest of the process,
+        # regardless of the stubbing method used.
         if @resource.try(:taggable?) && @resource.respond_to?(:acts_as_taggable)
           @resource.acts_as_taggable.each do |field|
             tagged_records = @records.tagged_with(@search.downcase)
             condition = acts_as_taggable_query(tagged_records)
-            conditions << condition if condition
+            push_condition(conditions, condition, @resource.primary_key.to_s) if condition
```
christophebrun-forest: Two things drift from the contract `push_condition` states ("the footprint reported to the permission guard can never drift from what the generated SQL actually reads"):

1. `acts_as_taggable_query` runs `.map { |t| t[@resource.primary_key] }` on `tagged_records`, which **executes** the query — an `INNER JOIN taggings … JOIN tags`. The path recorded here is `@resource.primary_key`, i.e. a root column, so neither `taggings` nor `tags` ever appears in the footprint.
2. That read happens inside `search_param`, so it is emitted *before* `assert_can_read_query_fields` runs (line 30). For this branch specifically, the SQL goes out ahead of any permission check.

In practice `tags`/`taggings` are rarely exposed Forest collections, so the permission impact is small. But "can never drift" isn't accurate as written, and this branch has no spec coverage (understood, and the reason is documented above).

Related, maybe worth settling in the same pass: line 35 filters the footprint with `.select { |path| path.include?(':') }`, so every root path recorded by `push_condition` is dropped before the check. The root-column recordings (lines 90-111) are therefore inert as far as permissions go — justified by the column/association name-collision comment, but it does mean the effective scope of the `push_condition` refactor is relation paths only. Worth a word in the comment so a future reader doesn't assume root columns are checked.

app/services/forest_liana/ability/permission.rb
```diff
@@ -159,6 +160,13 @@ def assert_can_read_query_fields(user, root_model, filter_paths: [], sort_paths:
         denied = usages.find { |usage| !FieldPath.readable_leaves?(usage[:collections], readable_collection_names) }
         return unless denied
 
+        unexposed = denied[:collections].reject { |name| collection_exposed?(name) }
```
christophebrun-forest: `unexposed.any?` short-circuits, and for a polymorphic path `denied[:collections]` holds several targets. If one target is unexposed and another is exposed but read-denied, only the exposure is reported: the admin exposes the named collection, replays the request, and now gets "read denied" on the other one. Two round trips for a diagnosis that was available in one — which cuts against the goal of this code, telling the caller *why* the request was refused.

Suggestion: carry both lists (unexposed / not readable) in the raised error, or at least mention in the message that other targets on the same path are also involved.

Same family, line 160: `usages.find` keeps only the first failing usage, so a request that filters on a denied collection *and* sorts on another only ever reports the filter. Probably intentional (one refusal is enough to reject the request), so only worth revisiting if a complete diagnosis in a single round trip is a goal.

christophebrun-forest (commented):
Two non-blocking findings, both about a claim being slightly stronger than the code (comment accuracy / error message completeness) rather than about behaviour.

christophebrun-forest:
## Consolidating the three plain-search / extended-search findings

The three issues below look separate but come from one place: this PR introduces the invariant *"a search that cannot be resolved must never fall through to the unfiltered table"*, implements it at **two sites with two different notions of failure**, and reuses the predicate that serves it (`smart_search_declared?`) for a **second, unrelated invariant** — "a search whose footprint cannot be described must not be served".

### The three symptoms

**(1) `@records.none` on a lambda failure — `search_query_builder.rb:56`.** At that point `@records` already carries the column conditions built by `search_param` plus the scopes and filters applied above. A smart-field `search:` lambda that raises (typically an external call timing out) therefore discards work that succeeded: a search for "Oak" on a collection with text columns returns 0 rows instead of the 5 `name LIKE '%oak%'` matches, with a 200 and an empty list. And because the loop keeps going with `@records` now a `NullRelation`, any later lambda's contribution is unreachable — one broken lambda silences the whole search.

**(2) The describability guard is gated on `searchExtended` — `search_query_builder.rb:331`.** The comment justifies exempting a plain search on the grounds that everything it reads besides the lambda is root-only and pinned readable. That is true of the *columns*, but the leak the error names is the lambda itself ("a smart-field search lambda can read anything"), and the lambda runs identically on a plain search (lines 45-60) with its results served. Since `searchExtended` is a caller-controlled query param, anyone wanting the lambda's reach simply drops it. Net effect: every customer using the documented `search:` hook loses extended search (4xx), while the arbitrary read stays served on the default path — the cost is paid without the benefit.

**(3) `!smart_search_declared?` empties plain searches on dotted-only `search_fields` — `search_query_builder.rb:193`.** For a collection declaring e.g. `search_fields: ['author.name', 'category.label']` and no smart search: no root column is in `search_fields`, so the column loop builds nothing; the dotted branches sit under `if extended_search?`, so they never run; `conditions.empty?` + no lambda → `none`. Every plain search returns 0 records for any term. Note `compute_includes` (`resources_getter.rb:114-118`) builds `includes_for_smart_search` from exactly those dotted entries **in the non-extended branch** — so the getter joins the relations that the builder then refuses to search. To be fair, pre-PR this case returned the *whole unfiltered table*, which is worse; but it is still a visible behaviour change outside the PR's stated scope, and pinned by no spec.

### Where it actually goes wrong

| Site | Notion of failure | Invariant served | Tool |
|---|---|---|---|
| 193 | `conditions.empty?` + `!smart_search_declared?` | coverage | `@resource.none` |
| 56 | the lambda raised | coverage | `@records.none` |
| 331 | a lambda is declared **and** `searchExtended=1` | describability | 4xx refusal |

Two structural problems fall out:

- **The coverage decision is taken too early (193) and retaken too late (56).** At 193 we don't yet know whether the lambda will contribute, so `smart_search_declared?` stands in for "something will constrain the query". At 56 we no longer know what the columns already produced, so `.none` overreaches. Each site compensates for the other's blind spot, and each is wrong in one case.
- **One predicate answers two orthogonal questions.** "Will anything constrain the query?" (functional, decides *what* we serve) and "can I describe what is read?" (security, decides *whether* we serve). No single predicate can be right for both.

### Proposed consolidation

Decide each axis once, where the information is complete. `push_condition` is already the single funnel for every condition, so it can carry the coverage flag:

```ruby
def perform(resource)
  @resource = @records = resource
  @tables_associated_to_relations_name = ...
  @search_constrained = false

  assert_search_describable!    # axis B — no longer gated on searchExtended
  @records = search_param       # no longer emits `none` itself

  assert_can_read_query_fields(@user, root_model, filter_paths: ..., search_paths: ...)
  # scopes / filters, unchanged

  apply_smart_search!           # the lambda loop, extracted

  @records = @records.none if @search && (malformed_uuid_search? || !@search_constrained)

  @records = sort_query
  @records
end

def push_condition(conditions, condition, path)
  return unless condition

  @search_field_paths << path
  @search_constrained = true
  conditions << condition
end
```

and in the lambda loop:

```ruby
begin
  @records = field[:search].call(@records, @search)
  @search_constrained = true
  (@fields_searched << field[:field].to_s) if field[:type] == 'String'
rescue => exception
  FOREST_REPORTER.report exception
  FOREST_LOGGER.error "Cannot search properly on Smart Field:\n#{exception}"
  # nothing else: a failed lambda simply doesn't count as a contributor
end
```

`malformed_uuid_search?` stays an independent, unconditional emptying — its rationale (LIKE scans hitting the statement timeout, and never serving the whole table for a mistyped UUID) holds even when a lambda ran, so it must not be folded into the flag.

One assumption to state out loud: "a lambda that ran without raising counts as a contributor" is a proxy — a lambda returning `query` untouched constrains nothing but sets the flag. That is exactly the assumption line 193 already makes today via `smart_search_declared?`, so this inherits it rather than introducing it.

### What this does to the existing specs

Everything currently pinned in `search_query_builder_spec.rb` stays green, which is the main reason I think this is the low-risk shape:

- *"answers no records rather than the whole table"* (`search_fields: ['nonexistent']`, no lambda) → no condition, no lambda, flag false → `none`. ✅
- *"is left unfiltered, since the lambda ORs its own conditions in afterwards"* (lambda returns `query`) → flag true → no `none`, no `WHERE`. ✅
- *"answers no records rather than silently falling through to the unfiltered table"* (lambda raises) — note this case sits inside the `search_fields: ['nonexistent']` block, so no column matched either → flag false → `none`. ✅
- *"a blank or whitespace-only search … is served"* → `@search` nil, guard short-circuits. ✅
- The malformed-UUID contexts → preserved by the explicit term. ✅

Only **one spec needs adding**, and it is precisely the regression: a collection with matching text columns *and* a lambda that raises must still serve the column matches. The current code returns 0 there and nothing pins it.

### The two decisions this doesn't make for you

- **Finding 2** is a design call, not a bug to patch: either the lambda is undescribable and the refusal applies in **both** modes (consistent, but breaks every customer of the hook), or it is acceptable and there is no refusal, with a documented note that smart-field lambdas are out of the permission footprint by construction. My preference is the second — the current middle ground closes a door while leaving the window open. Either way, dropping `extended_search?` from the guard's condition is what makes the choice explicit.
- **Finding 3** reduces to one explicit question: does a plain search cover dotted `search_fields`?
  - **(b), my suggestion** — no, extended-only: then also drop `includes_for_smart_search` from the non-extended branch of `compute_includes` (a join paid for a search that never happens; worth confirming nothing else relies on it) and add a spec pinning the `none`.
  - **(a)** — yes, build those conditions on plain searches too: aligned with what the getter already joins, but it widens search semantics for every customer *and* adds those relation paths to the checked footprint. Out of this PR's scope, in my view.

app/services/forest_liana/search_query_builder.rb
matthv: Fixed in 4163373 — switched acts_as_taggable_query to a subquery (.to_sql instead of .map), deferring execution past assert_can_read_query_fields. Also corrected the comment's overclaim and noted the root-path exclusion where you flagged it.

app/services/forest_liana/ability/permission.rb
matthv: Fixed in 9692410 — UnexposedQueryCollectionError now also carries the exposed-but-denied targets alongside the unexposed ones, in one message/data payload.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -274,5 +307,33 @@ def sort_field_path(field)
...
+      ForestLiana.schema_for_resource(root_model)&.fields&.any? { |field| field.try(:[], :search) }
+    end
+
+    # A smart-field search lambda can read anything, so an extended search on a collection that
+    # declares one has no footprint to check against permissions — refused before anything runs,
+    # rather than left to compare a partial footprint. A plain search on the same collection is
+    # unaffected: what it reads besides the lambda is root-only and pinned readable, so nothing
+    # checkable is skipped by serving it.
+    def assert_extended_search_describable!
+      return unless @search && extended_search? && smart_search_declared? && has_permission_system?
```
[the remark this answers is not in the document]
matthv: Same underlying issue as Christophe's consolidated review comment (finding 2) — the guard being gated on searchExtended is one of two explicit options he laid out (refuse in both modes vs. never refuse). Went with the latter: smart-field search lambdas are now documented as outside the checked footprint by construction, in both modes, rather than gated on searchExtended at all. Rationale: the lambda runs identically regardless of that param, so a caller wanting its reach simply omits it — the gated refusal was costing every customer of the search: hook their extended search without closing the actual leak. assert_extended_search_describable!/smart_search_declared? are removed.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -74,37 +91,44 @@ def search_param
...
+        # a live gap. Not covered by a live spec — acts_as_taggable_on isn't installed in the dummy
+        # app, and stubbing `taggable?`/`acts_as_taggable` on the real Tree model to simulate it
+        # permanently corrupts ActiveRecord::Delegation's per-class method cache for the rest of
+        # the process, regardless of the stubbing method used.
         if @resource.try(:taggable?) && @resource.respond_to?(:acts_as_taggable)
           @resource.acts_as_taggable.each do |field|
             tagged_records = @records.tagged_with(@search.downcase)
-            condition = acts_as_taggable_query(tagged_records)
-            conditions << condition if condition
+            push_condition(conditions, acts_as_taggable_query(tagged_records), @resource.primary_key.to_s)
```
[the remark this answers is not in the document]
matthv: Already the accepted tradeoff the comment right above this call documents — acts_as_taggable_on's own tables are rarely exposed as Forest collections themselves in practice, which is what keeps this from being a live gap; authorizing them isn't possible today since they aren't a Forest relation this class can name/check. Not something this PR is changing.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -36,6 +52,10 @@ def perform(resource)
               FOREST_REPORTER.report exception
               FOREST_LOGGER.error "Cannot search properly on Smart Field:\n" \
                 "#{exception}"
+              # A failed lambda is the same "cannot be evaluated" case as no column matching at
+              # all: answering no records keeps the guarantee this file makes elsewhere, that a
+              # search that cannot be resolved never falls through to the unfiltered table.
+              @records = @records.none
```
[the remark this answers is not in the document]
matthv: Fixed in cb33793 — as part of a broader consolidation (a single @search_constrained flag decided once, after both the columns and every lambda have run, rather than @records mutated mid-loop). A later lambda succeeding can no longer un-empty an earlier one's failure, since the flag only ever gets set true, never reset.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -48,17 +68,18 @@ def format_column_name(table_name, column_name)
...
+    # condition — no behavior change for a search that tags nothing.
     def acts_as_taggable_query(tagged_records)
-      ids = tagged_records
-        .map {|t| t[@resource.primary_key]}
-        .join(',')
-
-      if ids.present?
-        return "#{@resource.primary_key} IN (#{ids})"
-      end
+      "#{@resource.primary_key} IN (#{tagged_records.select(@resource.primary_key).to_sql})"
```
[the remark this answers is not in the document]
matthv: Fixed in 1ad71d7 — qualified the primary key with the resource's own table name.

app/services/forest_liana/search_query_builder.rb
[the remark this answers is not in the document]
matthv: No, thanks — this is a deliberate, already-made decision, not an oversight. The read a smart-field lambda can reach was never inside the checked footprint to begin with (it isn't recorded in search_field_paths on any search, extended or plain); the searchExtended-gated refusal only ever closed that door in one of the two modes it's reachable from, at the cost of breaking every legitimate use of the search: hook whenever a request happened to set that flag. Removing the refusal doesn't create the gap — it was already there on the plain path — it just stops the inconsistent partial mitigation and documents the actual boundary. If this needs to be closed for real, that's a separate, larger change (authorizing or statically describing arbitrary lambda code) than this PR's scope.

matthv:
## Addressed — consolidation implemented, both open decisions made

Went with the consolidated shape you proposed. Concretely:

**Finding 1 (a raise discarding already-found matches) — fixed.** A single `@search_constrained` flag, set by `push_condition` (already the one funnel every column/tag/association condition goes through) and by a smart-search lambda that runs without raising, decided once in `#perform` after both `search_param` and every lambda have run. A failed lambda no longer forces `@records.none` itself — it just doesn't count as a contributor, so an already-found column match (or a later lambda's own success) survives it. New spec pins exactly this: a real column match + a raising lambda → the column match is still served.

**Finding 2 (describability gated on searchExtended) — went with your preference: no refusal, in either mode.** `assert_extended_search_describable!`/`smart_search_declared?`/`UndescribableSearchError` are removed entirely, replaced with a comment documenting that a smart-search lambda's reach is outside the checked footprint by construction — it runs identically regardless of `searchExtended`, so gating a refusal on that param was only ever costing every customer of the hook their extended search without closing the actual read. (Macroscope pushed back on this exact point after the removal — replied there reaffirming it's deliberate, not an oversight; happy to revisit if you land somewhere else on it.)

**Finding 3 (dotted-only search_fields empties plain search) — went with extended-only.** The dotted association conditions were already gated on `extended_search?` in `search_param`, so nothing changed there; what did need fixing was `resources_getter#compute_includes`, which still eager-loaded that association on a *plain* search — a join paid for a search that never ran. Dropped it from the non-extended branch, and added a spec pinning the `none` a collection with only a dotted search_fields entry now serves on a plain search (previously accidental, now asserted).

**Also caught and fixed in this same round** (by a full review pass after the consolidation, not part of your three):
- A real regression the consolidation itself introduced: a malformed-UUID-shaped search that also happened to match a real id/enum/tag (a numeric-leading string, say) was being emptied unconditionally, where it used to be served — `malformed_uuid_search?` only ever needed to suppress the LIKE-scan branches (which already guard on it individually), so it was redundant with `@search_constrained` for the "nothing matched" case and wrong for the "something else matched" one.
- A critical, unrelated bug in `acts_as_taggable_query` (from the earlier `.to_sql`-subquery fix): its subquery text was joined into the same string passed to `where(sql, binds)` as every other condition, so a tag name containing a colon (`note:urgent`) crashed the whole search with `PreparedStatementInvalid` — Rails' named-bind scanner has no notion of "this colon is inside an already-quoted SQL literal, not a placeholder". Fixed by substituting binds into the non-tag conditions first, then joining the tag condition in after — nothing bind-scannable survives to that point. Also swapped `.select` for `.reselect` on the tag subquery (a query builder that already selects its own columns would otherwise produce a subquery `IN` can't use) and qualified both sides of the `IN`, not just one.

Commits: `cb337932`, `1ad71d79`, `92b89d7c`, `1a876816`, `59bc0b5`. 761 examples, 0 failures; every fix above has its own mutation check.

christophebrun-forest:
## Follow-up round: one regression from the consolidation, plus two smaller things

Context first, since it changes how to read the rest: I merged `origin/feature/optim-rbac-capabilities` into this branch (clean, no conflict) and ran the suite — **769 examples, 0 failures**. I also mutation-checked four of the new guards (`reselect`→`select`, dropping the `model_included?` filter on `many_associations`, dropping the `.none`, `search_paths: []`): each one turns specs red. The tests are load-bearing. The branch does need catching up though — merge-base is `5a9542c`, the base is at `23e6803`.

---

### 1. `92b89d7` over-corrected: a malformed-UUID search now serves the whole table when a smart-search lambda is declared

The divergence from the consolidated shape is a single term. What I proposed, and what `cb33793` implemented:

```ruby
@records = @records.none if @search && (malformed_uuid_search? || !@search_constrained)
```

with the note that `malformed_uuid_search?` had to stay an independent term. `92b89d7` removed it:

```ruby
@records = @records.none unless @search_constrained
```

**You were right to remove it as written.** My form was too strong: a UUID-shaped string starting with digits goes through `@search.to_i` and pushes a real `trees.id = …` condition, and the unconditional `malformed_uuid_search?` emptied it anyway. Putting my form back on the current branch fails exactly the spec you added:

```
770 examples, 1 failure
  1) …#perform when the malformed UUID also happens to match a real id still serves the id match
```

My mistake was writing `||` instead of scoping the term. The base rule was `none if conditions.empty? && malformed_uuid_search?` — the malformed term only ever mattered *when `search_param` had produced nothing*.

**But removing it entirely misses why it was there.** In the base, the `.none` was emitted inside `search_param`, **before** the lambda loop, so a lambda could never undo it — a no-op lambda received a `NullRelation` and handed it straight back. Moving the decision after the loop is the right call, but it hands the lambda a veto it never had.

`@search_constrained` conflates two different contributors: "`push_condition` pushed a condition" and "a lambda didn't raise". Only the first is actually a constraint — a lambda that returns its relation untouched sets the flag without filtering anything. `92b89d7`'s own message enumerates the counterexample and then doesn't apply it: *"that flag can only be true here via an id/enum/tag match **or a lambda**"* — the lambda is precisely the case where the term wasn't redundant.

Measured record counts, 2-row table, plain (non-extended) search:

| case | base | current | my proposal | two flags |
|---|---|---|---|---|
| **A.** malformed UUID that also matches a real id | 1 | 1 ✅ | **0** ❌ | 1 ✅ |
| **B.** malformed UUID, no column match, no lambda | 0 | 0 ✅ | 0 ✅ | 0 ✅ |
| **C.** malformed UUID, no column match, no-op lambda | 0 | **2** ❌ | 0 ✅ | 0 ✅ |
| **D.** malformed UUID, no column match, constraining lambda | 0 | 1 | 0 | 0 |
| **E.** ordinary search, no column match, no-op lambda | 2 | 2 ✅ | 2 ✅ | 2 ✅ |

Each of us fixed the case the other missed. Separating the two contributors covers both:

```ruby
# perform
@conditions_pushed  = false
@lambda_contributed = false
…
  # after the lambda loop
  @records = @records.none unless @conditions_pushed || (@lambda_contributed && !malformed_uuid_search?)

# push_condition : @conditions_pushed  = true
# lambda loop    : @lambda_contributed = true
```

**770 examples, 0 failures**, and A–E back to the base's behaviour including your `92b89d7` spec.

Case **D** stays an explicit call: the base answered 0, the current code answers 1 (a lambda that really filters gets served despite the UUID being malformed). Defensible, but it isn't what `92b89d7`'s message describes changing, and nothing pins it either way.

#### Why no mutation check caught this

Three of the specs covering this decision can't fail on what they name:

```
Tree.none.to_sql  =>  ""
Tree.all.to_sql   =>  "SELECT \"trees\".* FROM \"trees\""
```

`expect(…to_sql).not_to match(/\bWHERE\b/i)` holds for an unfiltered relation **and** for a `NullRelation`. So:

- `search_query_builder_spec.rb:367` — *"is left unfiltered, since the lambda ORs its own conditions in afterwards"*
- `search_query_builder_spec.rb:409` — *"is served, since the later lambda still constrained the query"* (the one added in `92b89d7` to pin that the flag only ever moves false→true)
- `search_query_builder_spec.rb:319` — *"reports nothing, matching the `.none` it falls through to"*

all pass in both directions. I confirmed it by running a variant that flips case E from 2 records to 0: suite still green. `.count` / `.pluck` would bite here, `to_sql` doesn't — same tool the two neighbouring specs already use successfully.

---

### 2. The title and part of the description no longer describe the code

The title still reads *"…and refuse an extended search that cannot be described"*, and the **What** section still describes `UndescribableSearchError` ("an extended search on it is refused before anything runs"). That mechanism was removed in `cb33793`, and a spec now pins the opposite (200, never refused). Since the repo squash-merges, this title becomes the commit message and the changelog entry.

Same section, the **Review round** bullet: *"`resources#destroy_bulk` and `associations#dissociate` now return 403 instead of a bodiless 500"*. On the base branch both already reached `rescue ForestLiana::Errors::ExpectedError` → **403 with a body** (`status` + `detail`) — checked on `origin/feature/optim-rbac-capabilities`. The actual improvement is that the payload now also carries `name` and `data`, which is worth stating as such so nobody reads it as a crash that got fixed.

---

### 3. `error.display_error` is no longer called on those two routes

In `destroy_bulk` / `dissociate`, these errors previously fell into the `rescue ForestLiana::Errors::ExpectedError` branch, which calls `error.display_error` before rendering. The new `rescue *QUERY_PERMISSION_ERRORS; raise` routes them to `render_error`, which doesn't. `index`/`count` never logged them either, so the end state is consistent — just checking it's the intended side effect rather than an unnoticed one.

---

Nothing else on the security side: the footprint covers both extended-association loops, polymorphic is excluded from SQL and footprint alike, root-column paths are deliberately dropped with the root pinned readable upstream, and `sanitize_sql_array` → `where(String)` is correct — I reproduced the colon/quote/`?` case (`o'ak:urgent ?`) and it executes fine now.

matthv:
Thanks for the very thorough repro table — you were right on every count.

**1. Fixed** (`2f22a6b`) — split into `@conditions_pushed`/`@lambda_contributed` exactly as proposed, malformed_uuid_search? weighed against the lambda contributor alone. Case D (a genuinely-filtering lambda still losing to a malformed-UUID-shaped term) is accepted as you left it — a proper fix needs comparing the lambda's own before/after relation, out of scope for this regression fix. Made it observable instead (`c9e4f541`): logged when it fires, and pinned by a new spec using an actually-filtering lambda rather than the existing no-op one, so it can't silently flip back.

Also fixed the three mutation-blind specs — switched to `.count` against real seeded rows, and (a follow-up review round caught) made the seeded-row count itself part of the assertion, since the `eq(0)` ones were vacuous if their own `before` block ever drifted.

**2. Fixed** — title and description updated (base branch, the "What" section, the review-round summary, example count).

**3. Confirmed intentional** — `index`/`count` never called `display_error` for these either, so this makes `destroy_bulk`/`dissociate` consistent with them rather than less logged than before.

771 examples, 0 failures, pushed.

app/services/forest_liana/search_query_builder.rb
```diff
@@ -20,30 +20,80 @@ def perform(resource)
...
+              # already matched) must still be free to serve the request.
             end
           end
         end
+
+        # A real condition (id/enum/tag/column match) is served regardless of malformed_uuid_search?
+        # — it never suppressed those branches, only the LIKE scans. A lambda's own contribution is
+        # weighed against it instead: a malformed-UUID-shaped search is still emptied if all that
+        # "constrained" it was a lambda, the same guarantee the base gave before a lambda could run
+        # at all. Neither contributor at all is the one case left to fall through to the whole table.
```
christophebrun-forest: The last sentence says the opposite of the line it documents:

```ruby
@records = @records.none unless @conditions_pushed || (@lambda_contributed && !malformed_uuid_search?)
```

With neither contributor, `@conditions_pushed` is `false` and `(@lambda_contributed && …)` is `false` too, so `unless false` runs `.none` — the query is **emptied**, it does not fall through to the whole table. That's this PR's own headline behaviour change ("answers **no records**, not the whole table"), and the `answers no records rather than the whole table` spec pins it.

No runtime impact, but this is the comment documenting the security-relevant line of the whole change: someone realigning the code on the comment later would reintroduce exactly the leak being closed here. Maybe something like "every case is now either constrained or emptied — nothing falls through to the whole table any more", with the lambda-that-actually-ran case called out as the only one that survives a malformed-UUID term.

matthv: Confirmed, that comment said the opposite of what the code does — fixed: `60c62755`. Thanks for catching it.

christophebrun-forest (approved):
LGTM

PR #800, opened 2026-09-04
test(perf): pin the per-request SQL footprint of list, count and show, N+1s included

matthv:
Instrument ticket for the perf work (PRD-1086 lazy column loading, PRD-1089 smart-field relation preload, PRD-1090 disable-count/no-forced-joins) — none of those can show a real before/after without something asserting on the query count and shape first.

No longer stacked on #799 (PRD-1142) — this PR now targets `main` directly. Independent of the read-permission stack (#794-#797): the guards add no query when permissions are disabled, and this harness doesn't need their code.

## What

`ForestLiana::QueryCapture` (spec/helpers) subscribes to `sql.active_record` and derives a **per-row delta** from two runs of the same request (2 rows vs 10), never an absolute query count — so a pin survives an unrelated query appearing or disappearing elsewhere in the response, and a query-cache bug shows up as a fractional `Rational` instead of a misleading round number.

`spec/requests/query_footprint_spec.rb` pins 7 scenarios against real generated SQL:

| Scenario | Pin |
|---|---|
| List projecting a to-one relation | 1 join, 0 extra queries/row |
| List projecting a cross-database `belongs_to` | 1 join on the same-db relation, 0 on the cross-db one, **1 query/row (N+1)** |
| List projecting a smart field walking a to-many relation | 0 joins, **1 query/row (N+1)** |
| Searched list + its count | list: 0/row ; count: forces a join it doesn't otherwise need |
| Get-one | 1 join, 1 standalone query for the cross-db relation, every column loaded |
| List projecting a polymorphic relation | 1 query/row on Rails 6, batched to 0 from Rails 7 — confirmed on live 6.1 and 8.0 legs of #799's matrix, not just reasoned about |

Each pin has a mutation-check: reverting the corresponding line in `base_getter.rb`, `resources_getter.rb`, `serializer_factory.rb`, or the Rails-version gate turns the matching example red — including the ticket's own AC, a deliberately reintroduced N+1 in the `has_one`/`belongs_to` serializer path.

## Also fixed

A latent bug in `bootstrapper_spec.rb`, found while chasing a full-suite-only failure: three of its examples call the real, unstubbed `ForestLiana::Bootstrapper.new`, whose `create_factories` rebuilds every model's serializer class from scratch. A `field`/`has_many`/`belongs_to` smart-field DSL block only ever runs once (when the collection file first loads), so a later real rebuild silently drops it from the serializer — system-wide, for the rest of the process. Nothing caught this before because the only other smart-field spec (`cap_name`) stubs serialization entirely. Fixed by stubbing `SerializerFactory` in the four examples that don't assert on it.

## Verification

518 examples, 0 failures, stable across repeated runs and random order. Confirmed on both the 6.1 and 8.0 legs of the CI matrix.

fixes PRD-1093

> [!NOTE]
> ### Fix `QueryCapture` SQL matching and pin footprint of list, count, show specs
> - Fixes `selects_from` to match only the root SELECT's FROM table by stopping at the first FROM keyword and escaping the table name in the regex, preventing nested-query FROM clauses from being mistaken for the root table
> - Expands `join_count` to count INNER JOIN, LEFT OUTER JOIN, and unqualified JOIN occurrences for the requested table, with the table name escaped in the pattern
> - Adds `Footprint#added_queries` to expose queries added during the grown capture, and updates footprint spec assertions to report the measured delta and added SQL statements on failure
> - Moves the query-capture support module from helpers to support, stubs `SerializerFactory` in the Bootstrapper spec, and removes per-context database cleanup hooks from the footprint request specs
> - Risk: `selects_from` in [query_capture.rb](https://github.com/ForestAdmin/forest-rails/pull/800/files#diff-da8d3924b68e0094cdf125123fa4bc7ba23a6a65269d647b103341d14a786a14) no longer matches a table appearing only in a nested subquery's FROM clause; any caller depending on that behavior will see fewer matches
>
> 
> #### Changes since #800 opened
>
> - Fixed `ForestLiana::QueryCapture.select_pattern`, `ForestLiana::QueryCapture::Footprint.added_queries`, and `ForestLiana::QueryCapture.join_count` to properly escape table names using `Regexp.escape` before using them in regular expressions, preventing false positives when table names contain special regex characters [fb4b150]
> - Added `ForestLiana::QueryCapture::Footprint.delta_report` helper method to format per-row delta and the list of added queries for failure diagnostics [fb4b150]
> - Added test coverage for `ForestLiana::QueryCapture.selects_from`, `ForestLiana::QueryCapture::Footprint.added_queries`, and `ForestLiana::QueryCapture.join_count` to verify proper handling of special characters in table names and multiset query differences [fb4b150]
> - Replaced custom expectation failure lambdas with calls to `result.delta_report` in `spec/requests/query_footprint_spec.rb` and added `after` cleanup hooks to destroy `Driver` records in multi-database test scenarios [fb4b150]
> - Removed redundant per-example stubbing of `ForestLiana::SerializerFactory` in `spec/lib/forest_liana/bootstrapper_spec.rb` while retaining the global before stub [fb4b150]
> - Added table filtering to `ForestLiana::QueryCapture::Footprint.delta_report` and updated test failure messages to use table-scoped reporting [49ad990]
> - Relocated and expanded query capture test suite [49ad990]
> - Refactored `ForestLiana::QueryCapture::Footprint.added_queries` control flow and expanded `footprint` helper documentation [49ad990]
> - Removed `Driver.destroy_all` after hooks from query footprint tests [49ad990]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 107b008.</sup>
> 
>


spec/lib/forest_liana/bootstrapper_spec.rb
```diff
@@ -38,7 +38,15 @@ module ForestLiana
         ]
       end
 
+      # A real SerializerFactory rebuilds every model's serializer class from scratch, which
+      # discards the smart-field attributes any Forest::* collection file attached to the
+      # previous generation (that DSL only runs once, when the file is first loaded) — so any
+      # example not asserting on serializers still stubs the factory to avoid corrupting the
+      # smart fields every other spec in the run relies on.
```
christophebrun-forest: **This names the wrong cause — and the fix should live in one place.**

The DSL doesn't "only run once". `Bootstrapper#generate_apimap` calls `require_lib_forest_liana`, which `load`s every collection file and re-attaches the smart fields to the freshly built serializer. The actual trigger is this describe's own `before` (line 4), stubbing `env_secret` to `nil`: `generate_apimap` is then skipped, so `create_factories` rebuilds the serializers and nothing ever re-attaches. Measured with a real `Bootstrapper.new` either way:

```
env_secret nil -> cap_name present? false
env_secret set -> cap_name present? true
```

So there's no production bug here, and the comment as written implies there is — which is the kind of note that sends the next person hunting in `collection.rb`.

Since the trigger is the `env_secret` stub, the mitigation belongs next to it rather than repeated in three examples; otherwise the landmine is back for the next example that builds a real `Bootstrapper`:

```ruby
before do
  allow(ForestLiana).to receive(:env_secret).and_return(nil)
  allow(ForestLiana::SerializerFactory).to receive(:new)
    .and_return(instance_double(ForestLiana::SerializerFactory, serializer_for: nil))
end
```

Verified: `bootstrapper_spec` + `projection_header_spec` in defined order → 41 examples, 0 failures, including `should generate serializers for all models`, which stubs the factory itself and still asserts on it.

spec/helpers/forest_liana/query_capture.rb
```diff
@@ -0,0 +1,56 @@
+module ForestLiana
+  module QueryCapture
+    NOISE_NAMES = %w[SCHEMA TRANSACTION].freeze
+    NOISE_SQL = /\A\s*(begin|commit|rollback|savepoint|release|pragma)\b/i
+
+    def self.select_pattern(table)
+      /\ASELECT\b.*\bFROM "#{table}"/im
```
christophebrun-forest: **This also matches a subquery, so `selects_from` can misattribute.**

`.` spans newlines under `/m`, so the pattern matches `FROM "table"` *anywhere* in the statement, not just in the root FROM clause:

```
SELECT "trees".* FROM "trees" WHERE "trees"."owner_id" IN (SELECT "users"."id" FROM "users" …)
→ selects_from(q, 'users') == [q]
```

None of the six scenarios produces a subquery, so it's green today. But the first pin that goes through a scope, a segment, or a `COUNT(DISTINCT …) FROM (SELECT …)` will get a wrong `selects_from` / `per_row_delta(table:)` — and it'll read as an N+1 that isn't there. Anchoring on the root FROM (and `Regexp.escape(table)` while you're in here) costs a line.

spec/helpers/forest_liana/query_capture.rb
```diff
@@ -0,0 +1,56 @@
...
+      seed.call(small)
+      baseline = capture_queries { yield small }
+      seed.call(large - small)
+      grown = capture_queries { yield large }
+
+      Footprint.new(baseline, grown, large - small)
+    end
+
+    def join_count(queries, table)
+      queries.sum { |sql| sql.scan(%(LEFT OUTER JOIN "#{table}")).size }
```
christophebrun-forest: **Only LEFT OUTER JOIN is counted, which makes the `eq(0)` pins blind.**

Verified: `join_count(['… FROM "trees" INNER JOIN "users" ON …'], 'users') == 0`.

The `eq(1)` assertions still fail loudly if the code stops eager-loading. But the `eq(0)` ones — `drivers` in the cross-database scenario, `trees` in the smart-field one — are false negatives: a regression switching to `joins` (INNER) keeps them green, and those two are exactly the assertions carrying the "never joins this" claim.

```ruby
queries.sum { |sql| sql.scan(/\b(?:LEFT OUTER |INNER )?JOIN "#{Regexp.escape(table)}"/).size }
```

spec/requests/query_footprint_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+    after { Tree.destroy_all; User.destroy_all }
+
+    it 'joins the relation once and reads nothing per row' do
+      result = footprint(seed: seed) do |rows|
+        get '/forest/Tree', params: params, headers: headers
+        expect(response).to have_http_status(200)
+        expect(listed_rows).to eq(rows)
+      end
+
+      expect(result.per_row_delta).to eq(0)
```
christophebrun-forest: **A broken pin hands you no SQL to work from.**

Breaking this expectation on purpose gives the whole message:

```
Failure/Error: expect(result.per_row_delta).to eq(5)
  expected: 5
       got: (0/1)
```

Whoever hits this on CI in six months has to re-run the spec by hand with a `puts` to find out which query appeared — on a harness whose entire job is to be read later by someone who didn't write it. `Footprint` already holds both sides, so a lazy message (or a small dedicated matcher) closes it:

```ruby
expect(result.per_row_delta).to eq(0), -> {
  "per-row delta #{result.per_row_delta}\nqueries added:\n#{(result.grown - result.baseline).join("\n")}"
}
```

spec/requests/query_footprint_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+          Tree.create!(name: 'tree', owner: user, cutter: user)
+        end
+      end
+    end
+    let(:params) do
+      { fields: { 'Tree' => 'id,name,owner', 'owner' => 'name' }, page: page,
+        searchExtended: '0', sort: '-id', timezone: 'Europe/Paris' }
+    end
+
+    after { Tree.destroy_all; User.destroy_all }
```
christophebrun-forest: **These `after` hooks are dead code — all six of them.**

`use_transactional_fixtures = true`, and the rollback covers the cross-database connections too, not only `primary`. Probed with two examples in one file:

```
PROBE-a: drivers=1 users=1
PROBE-b: drivers=0 users=0
```

No `before(:all)` in the suite creates a record either, so there's no leak for them to catch. It's inherited from `count_spec.rb`'s habit, but here six hooks suggest an isolation guarantee they aren't the ones providing.

spec/rails_helper.rb
```diff
@@ -6,6 +6,7 @@
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
+require_relative 'helpers/forest_liana/query_capture'
```
christophebrun-forest: **`spec/helpers/` is rspec-rails' directory for helper *specs*, not for shared support code.**

With `infer_spec_type_from_file_location!` on, everything under it is `type: :helper`, and it already holds `query_helper_spec.rb`. Requiring a non-spec module from there works (the default pattern only picks up `*_spec.rb`), but `spec/support/` — the directory documented in the comment block right below this line — is where the next reader will go looking for it.

christophebrun-forest (commented):
Reviewed by running the suite locally on Rails 6.1, plus a few targeted probes. Three things that don't anchor to a single line.

### 1. The base branch is `feature/optim-rbac-capabilities`, not `main`

The description says this PR "now targets `main` directly", but `baseRefName` is still the RBAC branch, which isn't merged into main yet. The *content* is genuinely independent — `git diff origin/main...718abd1` is the same 5 files / 314 insertions, and I ran the suite on `main` plus these five files: **519 examples, 0 failures**. So it's only the target to switch, but CI will need a re-run against the new base, since the current green checks were evaluated against the other one.

### 2. The pins do bite (for the record)

Mutating `eager_load` → `preload` in `base_getter.rb:37` turns **5 of the 7 examples red**, and the captured SQL matches the table in the description exactly: 10 × `SELECT "drivers".* … WHERE "id" = ?` for the cross-database list, and `SELECT COUNT(DISTINCT "trees"."id") … LEFT OUTER JOIN "users"` for the count.

The `bootstrapper_spec` fix is load-bearing too: with that file reverted, `bootstrapper_spec` + `projection_header_spec` in defined order gives **23 failures**; with it, **0**. That's a real class of order-dependent flakes removed — worth more than the "Also fixed" framing gives it.

### 3. Coverage gaps worth recording on the ticket

Not blocking — the PR does what it announces — but these are the paths the perf work will actually touch:

- **no pin on `/relationships/:assoc`** (`HasManyGetter`). That's the path behind #790, #792 and PRD-1034, and the one PRD-1089 (smart-field relation preload) is going to change. Without a before/after there, that ticket ships without a guard rail.
- **no pin with `searchExtended: '1'`**. All six scenarios are `'0'`, and extended search is precisely the forced-join path PRD-1090 targets.
- **no pin for a filter or a sort on a relation**, the other generator of a join nobody asked for.

spec/lib/forest_liana/bootstrapper_spec.rb
matthv: Confirmed — moved the SerializerFactory stub into this describe's own before, next to env_secret, and reworded the comment to name the real cause. Fixed: 8abc818d.

spec/helpers/forest_liana/query_capture.rb
matthv: Confirmed — anchored on the root FROM (negative lookahead stopping at the first FROM found) and added Regexp.escape. Added a spec that reproduces your exact subquery example and pins it, mutation-checked. Fixed: 8f62890e.

spec/helpers/forest_liana/query_capture.rb
matthv: Confirmed — now counts INNER JOIN too, with your exact suggested pattern. Pinned in the same new spec, mutation-checked. Fixed: 8f62890e.

spec/requests/query_footprint_spec.rb
matthv: Added a lazy failure message (queries added) to every per_row_delta assertion in the file. Fixed: 8f62890e.

spec/requests/query_footprint_spec.rb
matthv: Confirmed dead — reproduced your probe (drivers=0 users=0 with the hooks removed) before deleting all six. Fixed: 8f62890e.

spec/rails_helper.rb
matthv: Moved to spec/support/forest_liana/ and updated the require. Fixed: 8f62890e.

spec/support/forest_liana/query_capture_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        expect(dummy.selects_from([sql], 'users')).to eq([sql])
+      end
+
+      it "escapes the table name, rather than treat it as part of the regex" do
+        sql = 'SELECT "us.ers".* FROM "us.ers"'
+
+        expect(dummy.selects_from([sql], 'us.ers')).to eq([sql])
+        expect(dummy.selects_from([sql], 'usXers')).to be_empty
+      end
```
christophebrun-forest, line 29: **These two assertions don't test the escape — verified inert.**

Removing `Regexp.escape` from *both* call sites in `query_capture.rb` leaves this file at **10 examples, 0 failures**.

The reason is structural: the pattern is built from `'us.ers'` and matched against SQL that literally contains `"us.ers"`, which matches with or without escaping. The assertion needs the other direction — a pattern built from a metacharacter table must *not* match SQL where the `.` did its wildcard job:

```ruby
expect(dummy.selects_from([sql], 'us.ers')).to eq([sql])
expect(dummy.selects_from(['SELECT "usXers".* FROM "usXers"'], 'us.ers')).to be_empty
```

Mutation-checked: with `Regexp.escape` → 10/0; without → this example and the `join_count` one fail, and nothing else.

Same class of defect as the inert `(?:LEFT OUTER |INNER )?` group you caught yourself in c3ab9d44.

spec/support/forest_liana/query_capture_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        expect(dummy.join_count([one_join, two_joins], 'users')).to eq(3)
+      end
+
+      it "escapes the table name, rather than treat it as part of the regex" do
+        sql = 'SELECT "us.ers".* FROM "us.ers" INNER JOIN "us.ers" "u2" ON 1 = 1'
+
+        expect(dummy.join_count([sql], 'us.ers')).to eq(1)
+        expect(dummy.join_count([sql], 'usXers')).to eq(0)
+      end
```
christophebrun-forest, line 116: **Same here — passes with or without `Regexp.escape`.**

`join_count(sql, 'usXers')` returns 0 either way, since the SQL contains `"us.ers"`. The direction that discriminates:

```ruby
expect(dummy.join_count([sql], 'us.ers')).to eq(1)
expect(dummy.join_count(['SELECT "x".* FROM "x" INNER JOIN "usXers" "u2" ON 1 = 1'], 'us.ers')).to eq(0)
```

Unescaped that second one returns 1, escaped it returns 0 — verified.

spec/support/forest_liana/query_capture_spec.rb
```diff
@@ -0,0 +1,79 @@
+module ForestLiana
```
christophebrun-forest, line 1: **`spec/services/forest_liana/` contradicts the argument that moved the module out of `spec/helpers/`.**

The move was made on the right reasoning: that directory is rspec-rails' home for helper *specs*, and the next reader will look in `spec/support/`. By the same reasoning, the spec for a support module doesn't belong in `spec/services/forest_liana/`, which holds the specs for `lib/forest_liana/services/*` — grepping there suggests production code, and there is no `ForestLiana::QueryCapture` in `lib/`.

Co-locating it as `spec/support/forest_liana/query_capture_spec.rb` is the consistent spot: the default `spec/**/*_spec.rb` pattern picks it up, and the `Dir[…support…]` line in `rails_helper.rb` is commented out, so there's no double-load.

(Worth knowing for later: if anyone uncomments that `Dir[]`, the `require_relative` becomes a double require and `Footprint = Struct.new` will warn *already initialized constant*.)

spec/support/forest_liana/query_capture.rb
```diff
@@ -15,6 +21,18 @@ def per_row_delta(table: nil)
...
+      # baseline, not just the newly added ones — exactly wrong for an N+1, whose whole signature
+      # is the same query repeated more times in grown than in baseline.
+      def added_queries
+        remaining = baseline.tally
+        grown.reject { |sql| remaining[sql].to_i.positive? && (remaining[sql] -= 1) }
+      end
+
+      def delta_report
+        "per-row delta #{per_row_delta}\nqueries added:\n#{added_queries.join("\n")}"
+      end
```
christophebrun-forest: **`delta_report` drops the `table:` dimension, so the most precise pins still fail blind.**

Every `expect(result.per_row_delta).to eq(n)` now carries the report, but the three per-table assertions — `per_row_delta(table: 'drivers')`, `(table: 'trees')`, `(table: 'users')` — don't, and those are exactly the ones saying *which* table is read per row. A break there still hands back the bare `expected: 1 / got: (0/1)` that motivated this method in the first place.

```ruby
def delta_report(table: nil)
  "per-row delta #{per_row_delta(table: table)}\nqueries added:\n#{added_queries.join("\n")}"
end
```

then attach it at those three call sites too. Closes the gap completely, for near-zero cost.

spec/support/forest_liana/query_capture.rb
```diff
@@ -15,6 +21,18 @@ def per_row_delta(table: nil)
...
         Rational(after.size - before.size, rows_added)
       end
+
+      # Array#- is a set difference: it would drop every occurrence of a query already present in
+      # baseline, not just the newly added ones — exactly wrong for an N+1, whose whole signature
+      # is the same query repeated more times in grown than in baseline.
+      def added_queries
+        remaining = baseline.tally
+        grown.reject { |sql| remaining[sql].to_i.positive? && (remaining[sql] -= 1) }
+      end
```
christophebrun-forest, line 36: **Correct, but it leans on `0` being truthy.**

The multiset diff is right — both new examples confirm it. The `reject` condition is an arithmetic expression though, and only works because `0` and `-1` are truthy in Ruby, so a reader has to stop on it. Same length, explicit:

```ruby
def added_queries
  remaining = baseline.tally
  grown.reject do |sql|
    next false unless remaining.fetch(sql, 0).positive?
    remaining[sql] -= 1
    true
  end
end
```

spec/support/forest_liana/query_capture.rb
```diff
@@ -15,6 +21,18 @@ def per_row_delta(table: nil)
 
         Rational(after.size - before.size, rows_added)
```
christophebrun-forest, line 22: **Nothing covers what the whole argument rests on: `per_row_delta`'s Rational contract, and `capture_queries`' filtering.**

The new spec pins `select_pattern`, `added_queries` and `join_count`. But the description's central promise — *"a query-cache bug shows up as a fractional `Rational` instead of a misleading round number"* — lives on this line and is untested, as is the `payload[:cached]` / `NOISE_*` filtering that makes the measurement honest in the first place. Two or three pure examples cover both:

```ruby
it 'reports a fraction when the delta is not a clean multiple of the rows added' do
  expect(Footprint.new(%w[a], %w[a b c], 2).per_row_delta).to eq(Rational(1, 2))
end
```

plus one pinning that `per_row_delta(table:)` counts only the queries reading that table.

Two smaller things on the same method: `footprint(small: 2, large: 2)` raises `ZeroDivisionError` on `Rational(x, 0)` — not reachable with the defaults, but a `rows_added.positive?` guard or a word in `footprint`'s comment would say so. And the `page` size of 10 is coupled to `large: 10` nowhere in writing: raising `large` breaks `listed_rows == rows` in a confusing way.

spec/requests/query_footprint_spec.rb
```diff
@@ -75,7 +74,12 @@ def listed_rows
         page: page, searchExtended: '0', sort: '-id', timezone: 'Europe/Paris' }
     end
 
-    after { Product.destroy_all; Manufacturer.destroy_all; Driver.destroy_all }
+    # The one cleanup kept: unlike a same-database model, whether transactional fixtures roll
+    # back the driver/user-role connection too isn't exercised anywhere else in this file, and
+    # this repo has already hit one Rails-version-specific multi-database config quirk (see the
+    # comment in spec/dummy/config/database.yml) — cheap insurance against a leak nothing else here
+    # would catch, on the one connection whose rollback behavior is genuinely less proven.
+    after { Driver.destroy_all }
```
christophebrun-forest: **Re-probed on `fb4b150`: still a no-op, and it costs five lines of comment to defend.**

With and without the hook, probed after the whole file runs:

```
with    after { Driver.destroy_all }: trees=0 users=0 owners=0 products=0 manufacturers=0 addresses=0 drivers=0 cars=0
without                             : trees=0 users=0 owners=0 products=0 manufacturers=0 addresses=0 drivers=0 cars=0
```

`setup_fixtures` wraps every already-established pool (`connection_pool_list`), and the dummy's `connects_to` models are eager-loaded at boot — that's the same mechanism from 6.1 through 8.x, not something that flips by version. The original probe was right; *"trusting an empirical spot-check alone was thin"* names an uncertainty, not a different mechanism.

Keeping the insurance is defensible, but then: one line of comment rather than five, and the text says **"The one** cleanup kept" while there are two of them (`describe 'a get-one'` carries a bare one). My preference is to drop both, and add one back naming the version if #799's matrix ever surfaces a real leak on 8.x.

spec/lib/forest_liana/bootstrapper_spec.rb
```diff
@@ -2,6 +2,13 @@ module ForestLiana
   describe Bootstrapper do
     before do
       allow(ForestLiana).to receive(:env_secret).and_return(nil)
+      # env_secret stubbed to nil above skips generate_apimap (and the require_lib_forest_liana it
+      # calls, which is what re-attaches a Forest::* collection file's smart fields to the fresh
+      # serializer classes create_factories is about to build) — so nothing here re-attaches them.
+      # Every example not asserting on the factory itself stubs it, to avoid corrupting the smart
+      # fields every other spec in the run relies on.
```
christophebrun-forest: **This sentence describes the previous, per-example approach.**

"Every example not asserting on the factory itself stubs it" now sits on a stub that applies to *all* of them. What it should say is that the one example asserting on the factory re-stubs it locally — which does work correctly, I checked.

christophebrun-forest (commented):
Second review pass — ran the branch locally on Rails 6.1 (sqlite) and mutation-checked each of the fixes.

### What holds up

Full suite on `fb4b150`: **786 examples, 0 failures** in defined order. The three `query_capture.rb` fixes are genuinely pinned — reverting the greedy `.*` fails 1 example, reverting `added_queries` to `grown - baseline` fails 2, reverting `join_count` to LEFT OUTER only fails 3. And every negative pin (`be_empty`, `join_count … eq(0)`) does have a positive counterpart on the same table inside its own scenario (`per_row_delta(table: 'drivers')`, `(table: 'trees')`), so none of them is vacuous today. Both substantive points from the last round are properly closed.

Six things inline, plus this one. Two I'd fix before merge: the `Regexp.escape` coverage (inline on `query_capture_spec.rb`) and the description below.

### The description still carries the cause that 8abc818d corrected

It says *"A `field`/`has_many`/`belongs_to` smart-field DSL block only ever runs once (when the collection file first loads), so a later real rebuild silently drops it from the serializer"*. That is the wrong cause, and replacing it was the entire point of 8abc818d: `env_secret` stubbed to nil → `generate_apimap` skipped → `require_lib_forest_liana` never re-attaches. The code comment and the description now contradict each other, and the description is what the next reader hits first.

Three other stale bits while you're in there:

- *"Fixed by stubbing `SerializerFactory` in the four examples that don't assert on it"* — it's a single global `before` since fb4b150.
- *"518 examples, 0 failures"* — 786 on the current base.
- *"this PR now targets `main` directly"* — the base is still `feature/optim-rbac-capabilities` (unchanged since my last round).

spec/support/forest_liana/query_capture.rb
matthv: Confirmed — moved the table: param back into delta_report and wired it at all three call sites. Fixed: 49ad9905.

spec/support/forest_liana/query_capture_spec.rb
matthv: Confirmed inert, both were built in the direction that doesn't discriminate — rewrote with your exact suggested direction, mutation-checked. Fixed: 49ad9905.

spec/support/forest_liana/query_capture_spec.rb
matthv: Same fix as above, same reason. Fixed: 49ad9905.

spec/support/forest_liana/query_capture_spec.rb
matthv: Confirmed — moved to spec/support/forest_liana/. Fixed: 49ad9905.

spec/support/forest_liana/query_capture.rb
matthv: Rewrote explicit, same behavior. Fixed: 49ad9905.

spec/support/forest_liana/query_capture.rb
matthv: Added both — the Rational contract and the table: filtering, plus capture_queries' cached/SCHEMA/TRANSACTION filtering you noted was also untested. All mutation-checked (my first attempt at the table: test passed even with table: fully ignored — fixed that too before landing it). Added the rows_added==0/page-size coupling notes as comments rather than a guard, since neither is reachable with the defaults. Fixed: 49ad9905.

spec/requests/query_footprint_spec.rb
matthv: Agreed, re-probed myself and confirmed no leak either way — removed both, and the misleading "the one cleanup kept" comment with them. Fixed: 49ad9905.

spec/lib/forest_liana/bootstrapper_spec.rb
matthv: Confirmed — reworded to say the factory-asserting example re-stubs it locally, not that every other example still does its own stubbing. Fixed: 49ad9905.

christophebrun-forest (approved):
Great for the fixes, ok for me !

PR #802, opened 2026-09-08
feat(count): let a collection disable its count, and stop forcing joins into a non-extended searched count

matthv:
## Stack

`feature/optim-rbac-capabilities` (integration) → #799 → #800 → this PR.

fixes PRD-1090

## Context

Every list costs a second `/count` call. v2 lets a collection opt out (`disableCount()` → `schema.countable = false`, answered with `{ meta: { count: 'deactivated' } }`); v1 had no declarative equivalent. On the most expensive path, a search also forced the count to join every eager-loaded relation, even when the search is not extended and never touches an associated table.

## What changed

1. **`fix(collection)`** — `collection :X, read_only: true` / `is_searchable:` never had any effect on an AR-backed collection: `SchemaAdapter` creates the apimap entry before the custom collection file's opts get applied. Fixed so both are actually applied. Behavior change: a collection whose `read_only:`/`is_searchable:` was silently ignored until now will start honoring it.
2. **`feat(count)`** — `collection :X, countable: false` responds `{"meta":{"count":"deactivated"}}` with zero SQL, on both `/count` and the related `/count`. `count_related` tests the **foreign** collection's `countable`, matching v2's `count-related.ts:22`.
3. **`perf(count)`** — a search's count only joins associated tables when `searchExtended: '1'`; predicates on associations don't exist otherwise. Exception: a smart field with a `search:` lambda keeps its joins (its footprint is unknown, same reasoning as PRD-1085).

## Decisions

- **`countable`/`is_countable` is not meant to be a first-class apimap field — but it does leak into the raw payload today, for every collection, not just an opted-out one.** Corrected from an earlier, inaccurate claim here that the server would reject an unknown apimap attribute: `Model::Collection` only includes `ActiveModel::Serialization` (not `Serializers::JSON`), so its `as_json` falls back to `Object#as_json` → `instance_values` (every ivar), and `Bootstrapper` sends `ForestLiana.apimap.as_json` unfiltered in dev. `search_fields` already leaks the exact same way today without incident, so the server does tolerate it in practice — but the parity intent (response shape, not a schema addition) stands regardless of that leak, not because of the reasoning originally given here.
- `read_only`/`is_searchable` fix folded into this PR rather than shipped separately, since it's the same underlying mechanism (`apply_collection_opts`).

## Not touched

- `HasManyGetter#count` already has its own `optimize_record_loading`, unaffected by point 3.
- The two other `@count_needs_includes = true` sites (association-field filters) are correct and untouched.
- `Forest::OwnersController`/`OwnerTreesController` manual overrides still work as before.

## Release note (semver)

Point 1 (`fix(collection)`) activates `read_only:`/`is_searchable:` for a collection that may have carried either option, inertly, for years (neither was read by any guard before this PR). Under this repo's angular commitlint preset, `fix:` ships as a **patch** release — a client upgrading a patch version can silently lose their create/edit/delete buttons on an affected collection, with nothing about it standing out in a routine changelog scan. Calling it out explicitly here since the commit type itself won't carry the warning: worth a dedicated line in the release notes for whichever version ships this, not just the auto-generated changelog entry.

## Verification

529 examples, 0 failures (Ruby 3.3.7). Mutation-checks done for all three changes (revert → red at the exact expected spot, restore → green).

> [!NOTE]
> ### Add `is_countable` collection option and stop joining associations on non-extended search counts
> - Adds `is_countable` to `ForestLiana::Model::Collection` (default true) and passes a `countable:` option from the collection DSL in [collection.rb](https://github.com/ForestAdmin/forest-rails/pull/802/files#diff-25800522f84001b9f45f8b8fd4204870b706a4cadd6305f3a811010f79189385) into the resolved apimap collection via `apply_collection_opts`.
> - `count` actions in [resources_controller.rb](https://github.com/ForestAdmin/forest-rails/pull/802/files#diff-be3235120644036caa6865b266de1f6d79eb2d247f996aeae24397e241822c27) and [associations_controller.rb](https://github.com/ForestAdmin/forest-rails/pull/802/files#diff-d87aaccd92079688766587630744649f7ba2d87e3a0864816d740b161c527054) now return a deactivated-count response early when `count_deactivated?` is true, skipping `ResourcesGetter`/`HasManyGetter`.
> - Changes `ResourcesGetter` count-include logic in [resources_getter.rb](https://github.com/ForestAdmin/forest-rails/pull/802/files#diff-de9538c4fb8386da065cc3f2d8e37030e67a693d8261634ed0153ded9ef60782) so association includes during count only happen when `searchExtended == 1` or the schema has a smart-field search lambda (`smart_search_fields?`).
> - Risk: plain (non-extended) searches no longer join associations for the count query in `ResourcesGetter#initialize`; search terms that previously matched only on related records will no longer match. Collections omitting `countable:` keep `is_countable` true and are unaffected.
>
> 
> #### Changes since #802 opened
>
> - Modified `ForestLiana::Collection::ClassMethods.collection` class method to store `:countable` option in `collection_opts` and apply collection options only to existing apimap entries [8fcf6bc]
> - Updated count deactivation checks in `ForestLiana::AssociationsController.count` action to call `deactivate_count_response` based solely on `count_deactivated?(@association.klass)` without polymorphic guard [8fcf6bc]
> - Refactored `ForestLiana::ApplicationController.count_deactivated?` method to use local variable for collection name before apimap lookup [8fcf6bc]
> - Reordered initialization in `ForestLiana::ResourcesGetter` constructor to compute `@collection_name` and `@collection` before `@count_needs_includes`, and reimplemented `smart_search_fields?` to use already-resolved collection [8fcf6bc]
> - Added test coverage in `spec/lib/forest_liana/collection_spec.rb` to verify no apimap entry is created for non-existent collections, and updated `spec/requests/query_footprint_spec.rb` to seed data before count endpoint test [8fcf6bc]
> - Updated inline comments in `ForestLiana::AssociationsController.count` action and `ForestLiana::ResourcesGetter` class [2058b7d]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 8fcf6bc. 4 files reviewed, 1 issue evaluated, 1 issue filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>lib/forest_liana/collection.rb — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 32](https://github.com/ForestAdmin/forest-rails/blob/8fcf6bc7481edc1d1c86e476632b6008de001542/lib/forest_liana/collection.rb#L32): `collection` now only applies its options to an existing apimap entry. For a legitimate smart collection (no ActiveRecord class), `smart_collection?` creates its serializer but there is no SchemaAdapter entry to find, so `collection :MySmartCollection, countable: false` never registers the collection or its countability setting. Consequently the collection disappears from the schema and its count is not deactivated unless an unrelated field/action/segment later happens to invoke `model`. <b>[ Out of scope (triage) ]</b>
> </details>
>
>
> </details>
>

### DIFF
--- a/app/controllers/forest_liana/application_controller.rb
+++ b/app/controllers/forest_liana/application_controller.rb
@@ -110,6 +110,11 @@ module ForestLiana
       render serializer: nil, json: { meta: { count: 'deactivated'} }
     end
 
+    def count_deactivated?(model)
+      ForestLiana.apimap.find { |collection| collection.name.to_s == ForestLiana.name_for(model) }
+        &.is_countable == false
+    end
+
     private
 
     def render_error(exception)
--- a/app/controllers/forest_liana/associations_controller.rb
+++ b/app/controllers/forest_liana/associations_controller.rb
@@ -32,6 +32,9 @@ module ForestLiana
       #         through and dereferencing a nil @association, which surfaced as
       #         a double-render / 500 rather than the intended 404.
       return if performed?
+      return deactivate_count_response if !SchemaUtils.polymorphic?(@association) &&
+        count_deactivated?(@association.klass)
+
       begin
         getter = HasManyGetter.new(@resource, @association, params, forest_user)
         getter.count
--- a/app/models/forest_liana/model/collection.rb
+++ b/app/models/forest_liana/model/collection.rb
@@ -5,7 +5,7 @@ class ForestLiana::Model::Collection
   extend ActiveModel::Naming
 
   attr_accessor :name, :fields, :actions, :segments, :only_for_relationships,
-    :is_virtual, :is_read_only, :is_searchable, :icon,
+    :is_virtual, :is_read_only, :is_searchable, :is_countable, :icon,
     :integration, :pagination_type, :search_fields,
     # TODO: Remove once lianas prior to 2.0.0 are not supported anymore.
     :name_old
@@ -24,6 +24,7 @@ class ForestLiana::Model::Collection
     @icon ||= nil
     @is_read_only ||= false
     @is_searchable = true if @is_searchable.nil?
+    @is_countable = true if @is_countable.nil?
     @only_for_relationships ||= false
     @pagination_type ||= "page"
     @search_fields ||= nil
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -6,7 +6,11 @@ module ForestLiana
       @resource = resource
       @params = params
       @user = forest_user
-      @count_needs_includes = @params[:search].present?
+      # SearchQueryBuilder only joins an association under searchExtended (:103) — a plain
+      # search never leaves @resource's own columns. A smart field's search: lambda is the
+      # exception: it can reach anywhere, so its footprint stays unknown.
+      @count_needs_includes = @params[:search].present? &&
+        (@params['searchExtended'].to_i == 1 || smart_search_fields?)
       @collection_name = ForestLiana.name_for(@resource)
       @collection = get_collection(@collection_name)
       @fields_to_serialize = get_fields_to_serialize
@@ -133,6 +137,10 @@ module ForestLiana
     end
 
     private
+
+    def smart_search_fields?
+      ForestLiana.schema_for_resource(@resource).fields.any? { |field| field[:search] }
+    end
 
     def get_fields_to_serialize
       @params.dig(:fields, @collection_name)&.split(',')&.map(&:to_sym) || []
--- a/lib/forest_liana/collection.rb
+++ b/lib/forest_liana/collection.rb
@@ -6,11 +6,15 @@ module ForestLiana::Collection
     attr_accessor :collection_name
     attr_accessor :is_read_only
     attr_accessor :is_searchable
+    attr_accessor :collection_opts
 
     def collection(collection_name, opts = {})
       self.collection_name = find_name(collection_name).to_s
       self.is_read_only = opts[:read_only] || false
       self.is_searchable = opts[:is_searchable] || false
+      # Only the keys actually passed — `model` (re)applies them each time it runs, since
+      # SchemaAdapter already creates a real AR-backed collection's entry before this loads.
+      self.collection_opts = opts.slice(:read_only, :is_searchable)
 
       # NOTICE: Creates dynamically the serializer if it's a Smart Collection.
       if smart_collection? &&
@@ -21,6 +25,8 @@ module ForestLiana::Collection
         ForestLiana::SerializerFactory.new(is_smart_collection: true)
           .serializer_for(self)
       end
+
+      model
     end
 
     def action(name, opts = {})
@@ -211,7 +217,16 @@ module ForestLiana::Collection
         ForestLiana.apimap << collection
       end
 
+      apply_collection_opts(collection)
+
       collection
+    end
+
+    def apply_collection_opts(collection)
+      return unless collection_opts
+
+      collection.is_read_only = collection_opts[:read_only] if collection_opts.key?(:read_only)
+      collection.is_searchable = collection_opts[:is_searchable] if collection_opts.key?(:is_searchable)
     end
 
     def active_record_class
--- a/lib/forest_liana/collection.rb
+++ b/lib/forest_liana/collection.rb
@@ -14,7 +14,7 @@ module ForestLiana::Collection
       self.is_searchable = opts[:is_searchable] || false
       # Only the keys actually passed — `model` (re)applies them each time it runs, since
       # SchemaAdapter already creates a real AR-backed collection's entry before this loads.
-      self.collection_opts = opts.slice(:read_only, :is_searchable)
+      self.collection_opts = opts.slice(:read_only, :is_searchable, :countable)
 
       # NOTICE: Creates dynamically the serializer if it's a Smart Collection.
       if smart_collection? &&
@@ -227,6 +227,7 @@ module ForestLiana::Collection
 
       collection.is_read_only = collection_opts[:read_only] if collection_opts.key?(:read_only)
       collection.is_searchable = collection_opts[:is_searchable] if collection_opts.key?(:is_searchable)
+      collection.is_countable = collection_opts[:countable] if collection_opts.key?(:countable)
     end
 
     def active_record_class
--- a/spec/requests/query_footprint_spec.rb
+++ b/spec/requests/query_footprint_spec.rb
@@ -1,0 +1,233 @@
+require 'rails_helper'
+
+describe 'SQL footprint of a front call', type: :request do
+  before do
+    allow(ForestLiana::IpWhitelist).to receive(:retrieve) { true }
+    allow(ForestLiana::IpWhitelist).to receive(:is_ip_whitelist_retrieved) { true }
+    allow(ForestLiana::IpWhitelist).to receive(:is_ip_valid) { true }
+    allow_any_instance_of(ForestLiana::Ability).to receive(:forest_authorize!) { true }
+    # forest_authorize! is bypassed above, so the field-read guards it would otherwise gate must be
+    # too, or they call the permissions API through the file-backed cache another file last wrote.
+    Rails.cache.write('forest.has_permission', false)
+    allow(ForestLiana::ScopeManager).to receive(:fetch_scopes)
+      .and_return({ 'scopes' => {}, 'team' => { 'id' => '1', 'name' => 'Operations' } })
+  end
+
+  token = JWT.encode({ id: 38, email: 'michael.kelso@that70.show', first_name: 'Michael',
+                        last_name: 'Kelso', team: 'Operations', rendering_id: 16,
+                        exp: Time.now.to_i + 2.weeks.to_i, permission_level: 'admin' },
+                      ForestLiana.auth_secret, 'HS256')
+  headers = { 'Accept' => 'application/json', 'Content-Type' => 'application/json',
+              'Authorization' => "Bearer #{token}" }
+  page = { 'number' => '1', 'size' => '10' }
+
+  def listed_rows
+    JSON.parse(response.body)['data'].size
+  end
+
+  describe 'a list projecting a to-one relation' do
+    let(:seed) do
+      lambda do |n|
+        n.times do
+          user = User.create!(name: 'owner')
+          Tree.create!(name: 'tree', owner: user, cutter: user)
+        end
+      end
+    end
+    let(:params) do
+      { fields: { 'Tree' => 'id,name,owner', 'owner' => 'name' }, page: page,
+        searchExtended: '0', sort: '-id', timezone: 'Europe/Paris' }
+    end
+
+    after { Tree.destroy_all; User.destroy_all }
+
+    it 'joins the relation once and reads nothing per row' do
+      result = footprint(seed: seed) do |rows|
+        get '/forest/Tree', params: params, headers: headers
+        expect(response).to have_http_status(200)
+        expect(listed_rows).to eq(rows)
+      end
+
+      expect(result.per_row_delta).to eq(0)
+      expect(selects_from(result.grown, 'trees').size).to eq(1)
+      expect(join_count(result.grown, 'users')).to eq(1)
+      expect(selects_from(result.grown, 'users')).to be_empty
+
+      sql = selects_from(result.grown, 'trees').first
+      expect(sql).to include(column_ref('trees', 'name'), column_ref('users', 'name'))
+      expect(sql).not_to include(column_ref('trees', 'age'), column_ref('users', 'title'))
+    end
+  end
+
+  describe 'a list projecting a cross-database relation' do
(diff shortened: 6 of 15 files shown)


app/models/forest_liana/model/collection.rb
```diff
@@ -5,7 +5,7 @@ class ForestLiana::Model::Collection
   extend ActiveModel::Naming
 
   attr_accessor :name, :fields, :actions, :segments, :only_for_relationships,
-    :is_virtual, :is_read_only, :is_searchable, :icon,
+    :is_virtual, :is_read_only, :is_searchable, :is_countable, :icon,
```
christophebrun-forest, line 8: **🔴 1 — `is_countable` part bien dans l'apimap envoyé au serveur (en dev)**

La description dit « No `countable` flag in the apimap […] The server would reject an unknown apimap attribute ». C'est faux en l'état. Exécuté dans l'app dummy :

```
as_json keys              = [... "is_searchable", "is_countable", "only_for_relationships", ...]
serialized attribute keys = [... :is_searchable, :is_countable, :only_for_relationships, ...]
```

La chaîne : cette classe n'inclut que `ActiveModel::Serialization` (pas `Serializers::JSON`), donc `as_json` retombe sur `Object#as_json` → `instance_values` → **toutes** les ivars. En dev, `bootstrapper.rb:84` fait `@collections_sent = ForestLiana.apimap.as_json` (non filtré), et `schema_serializer.rb:55` recopie chaque attribut tel quel dans le POST. `SchemaFileUpdater::KEYS_COLLECTION` ne protège que le fichier `.forestadmin-schema.json`, pas le payload.

Conséquence : **toutes** les collections (pas seulement celles en opt-out) gagnent `is_countable: true` dans l'apimap envoyé au boot en dev.

Nuance importante : `search_fields` fuit déjà exactement pareil aujourd'hui et rien n'a jamais cassé — donc le serveur tolère visiblement les attributs inconnus. Mais alors c'est la *justification* de la décision qui est fausse, et il faut trancher : soit corriger la description, soit filtrer (alimenter `@collections_sent` depuis la sortie filtrée de `SchemaFileUpdater`, ou définir un `as_json` explicite ici).

christophebrun-forest (commented):
Relu en entier, suite exécutée sur la branche mergée : **800 examples, 0 failures**. Les vérifications ci-dessous sont empiriques (worktree isolé + mutation-checks), pas déduites.

Le fond des trois changements tient. J'ai revalidé le raisonnement de `perf(count)` : la seule branche de `search_param` qui pousse un prédicat sur une table associée est bien sous `extended_search?`, la sous-requête `acts_as_taggable` est auto-suffisante, et `FiltersParser#apply_filters` ajoute ses propres `eager_load` — donc les filtres sur association restent corrects sans `@count_needs_includes`. L'ordre dans `ResourcesGetter#initialize` est correct aussi : `@count_needs_includes` est assigné **avant** `field_names_requested`/`compute_includes`, donc le `true` posé par `extract_associations_from_filter` n'est pas écrasé. L'exception du lambda `search:` est le bon appel.

7 points ci-dessous. Deux demandent une décision explicite avant merge (1 et 3), deux sont des correctifs d'une ligne (2 et 4).

spec/requests/query_footprint_spec.rb
```diff
@@ -152,9 +152,32 @@ def listed_rows
...
+      expect(join_count(result.grown, 'users')).to eq(2)
       expect(selects_from(result.grown, 'trees').first).to match(/COUNT\(DISTINCT/)
     end
+
+    it 'matches nothing on the same term when the search stays on the root column' do
+      get '/forest/Tree/count', params: params.merge(search: 'owner'), headers: headers
+
+      expect(response).to have_http_status(200)
+      expect(JSON.parse(response.body)['count']).to eq(0)
+    end
```
christophebrun-forest, line 182: **🔴 2 — Ce test ne teste rien**

Pas de `seed` : la table est vide, donc `count == 0` est vrai quoi qu'il arrive.

Mutation-check exécuté — en annulant la ligne de `resources_getter.rb` (`@count_needs_includes = @params[:search].present?`), **seul** le test `:146` passe rouge ; celui-ci reste vert :

```
9 examples, 1 failure
rspec ./spec/requests/query_footprint_spec.rb:146 # counts in one statement without a join it does not need
```

Correctif vérifié (passe, et le test devient probant) :

```ruby
it 'matches nothing on the same term when the search stays on the root column' do
  seed.call(3)
  get '/forest/Tree/count', params: params.merge(search: 'owner'), headers: headers
  ...
```

lib/forest_liana/collection.rb
```diff
@@ -211,9 +217,19 @@ def model
...
+      apply_collection_opts(collection)
+
       collection
     end
 
+    def apply_collection_opts(collection)
+      return unless collection_opts
+
+      collection.is_read_only = collection_opts[:read_only] if collection_opts.key?(:read_only)
+      collection.is_searchable = collection_opts[:is_searchable] if collection_opts.key?(:is_searchable)
```
christophebrun-forest, line 233: **🔴 3 — `fix(collection)` → release patch, pour un changement visible côté client**

`is_read_only` / `is_searchable` au niveau collection ne sont lus par **aucun** garde côté agent — grep sur `app/controllers` et `app/services/forest_liana/ability` : zéro occurrence. Ce sont des attributs purement apimap, consommés par le front.

Donc pour un client qui traîne un `collection :X, read_only: true` sans effet depuis des années, l'upgrade fait disparaître les boutons créer/éditer/supprimer de son UI. Et avec `.releaserc.js` (preset angular), `fix(collection):` sort en **patch**.

Ça mérite un footer `BREAKING CHANGE:` ou au minimum une note de release explicite — c'est le genre de changement qui génère un ticket support à J+1.

lib/forest_liana/collection.rb
```diff
@@ -21,6 +25,8 @@ def collection(collection_name, opts = {})
         ForestLiana::SerializerFactory.new(is_smart_collection: true)
           .serializer_for(self)
       end
+
+      model
```
christophebrun-forest: **🟠 4 — Cet appel crée des collections fantômes**

`model` est maintenant appelé inconditionnellement. Un fichier qui ne déclare qu'un `collection :Typo, read_only: true` (nom erroné, modèle renommé ou supprimé) crée désormais une entrée virtuelle vide dans l'apimap — qui part dans le schéma et apparaît comme collection vide dans l'UI.

Vérifié en ajoutant `collection :Treees, read_only: true` (typo volontaire) à l'app dummy :

| branche | résultat |
|---|---|
| base | `nil` |
| PR | `#<Collection @name="Treees" @is_virtual=true @fields=[]>` |

Avant la PR : rien du tout, faute d'appel à `model`.

Suggestion : ne pas créer via `model` ici, seulement (ré)appliquer sur une entrée déjà présente. Les smart collections continueront de créer la leur via `field`/`action`, comme avant :

```ruby
existing = ForestLiana.apimap.find { |o| o.name.to_s == self.collection_name }
apply_collection_opts(existing) if existing
```

app/controllers/forest_liana/associations_controller.rb
```diff
@@ -70,6 +70,9 @@ def count
       return if performed?
       # Authorized outside the begin, like index above, so a denial keeps its name/data.
       forest_authorize!('browse', forest_user, @association.klass)
+      return deactivate_count_response if !SchemaUtils.polymorphic?(@association) &&
+        count_deactivated?(@association.klass)
```
christophebrun-forest: **🟠 5 — La garde polymorphe est du code mort**

La ligne juste au-dessus déréférence déjà `@association.klass`. Vérifié en tapant `/forest/Address/:id/relationships/addressable/count` sur cette branche :

```
Polymorphic associations do not support computing the class.
```

…levé par le `forest_authorize!`, **avant** d'atteindre `!SchemaUtils.polymorphic?`. Et la route est bien atteignable : `find_association` ne renvoie 404 sur un `belongs_to` que pour `index` (`params[:action] == 'index'`), pas pour `count`.

Le 500 est pré-existant, mais la garde telle qu'écrite laisse croire qu'il est traité. Soit la retirer, soit — mieux — remonter le `return if SchemaUtils.polymorphic?(@association)` **avant** l'autorisation et corriger le 500 au passage.

app/controllers/forest_liana/application_controller.rb
```diff
@@ -123,6 +123,11 @@ def deactivate_count_response
       render serializer: nil, json: { meta: { count: 'deactivated'} }
     end
 
+    def count_deactivated?(model)
+      ForestLiana.apimap.find { |collection| collection.name.to_s == ForestLiana.name_for(model) }
+        &.is_countable == false
+    end
```
christophebrun-forest, line 129: **🟡 6 — Invariant de boucle dans un PR de perf**

`ForestLiana.name_for(model)` (qui fait un `gsub`) est recalculé à **chaque** itération du `find`. Sur un apimap à 500 collections, ça fait 500 `gsub` + allocations par requête `/count` — pour une fonction dont le rôle est justement d'éviter du travail.

```ruby
def count_deactivated?(model)
  name = ForestLiana.name_for(model)
  ForestLiana.apimap.find { |collection| collection.name.to_s == name }&.is_countable == false
end
```

Au passage : c'est la 3ᵉ implémentation du même « trouver la collection d'un modèle » (`BaseGetter#get_collection`, `Bootstrapper#get_collection`, `ForestLiana.schema_for_resource`). Un `ForestLiana.collection_for(model)` serait le bon moment.

app/services/forest_liana/resources_getter.rb
```diff
@@ -129,6 +133,10 @@ def includes_for_serialization
 
     private
 
+    def smart_search_fields?
+      ForestLiana.schema_for_resource(@resource).fields.any? { |field| field[:search] }
+    end
```
christophebrun-forest, line 142: **🟡 7 — C'est le lookup le plus cher du fichier, et il tourne aussi sur `/list`**

`ForestLiana.schema_for_resource` est **O(collections × modèles)** : il appelle `find_model_from_collection_name` (qui reparcourt `ForestLiana.models`) pour chaque entrée d'apimap. Et comme `&&` court-circuite dans l'autre sens, il s'exécute à chaque construction de `ResourcesGetter` sur une recherche non étendue — donc aussi sur `/list`, pas seulement sur `/count`.

`@collection` (ligne 15) est **la même instance**. Vérifié sur les 14 modèles du dummy : `schema_for_resource(m).object_id == get_collection(name_for(m)).object_id` pour tous, et aucun `nil`.

Il suffit de déplacer l'affectation après la résolution de `@collection` :

```ruby
@collection_name = ForestLiana.name_for(@resource)
@collection = get_collection(@collection_name)
@count_needs_includes = @params[:search].present? &&
  (@params['searchExtended'].to_i == 1 || @collection.fields.any? { |field| field[:search] })
```

Bonus : supprime le `NoMethodError` sur `nil.fields` si la collection n'est pas dans l'apimap.

app/models/forest_liana/model/collection.rb
matthv: Confirmed, and the description was wrong — corrected. It does leak into the raw apimap payload for every collection today (Model::Collection's as_json falls back to Object#as_json/instance_values, sent unfiltered by Bootstrapper in dev), and search_fields already leaks the same way without incident, so the server does tolerate it — but that's not the reasoning the description gave. Not filtering the payload itself for now, since that's a separate, broader change than this PR's scope.

spec/requests/query_footprint_spec.rb
matthv: Confirmed and fixed — seeded, matches your correction.

lib/forest_liana/collection.rb
matthv: Added an explicit release-note section to the PR description flagging this for whoever cuts the release, rather than amending the already-pushed commit's type/rebasing.

lib/forest_liana/collection.rb
matthv: Confirmed and fixed — collection(...) now only (re)configures an already-existing apimap entry, matching your suggested fix. New regression test added.

app/controllers/forest_liana/associations_controller.rb
matthv: Confirmed dead — removed. Went with removing it rather than fixing the underlying 500: that crash is shared with index/update/associate/dissociate above and below (all call @association.klass directly), so fixing it properly is a wider change than this PR's own scope.

app/controllers/forest_liana/application_controller.rb
matthv: Fixed — hoisted out of the find loop.

app/services/forest_liana/resources_getter.rb
matthv: Confirmed and fixed — reuses @collection (moved its assignment earlier in the constructor) instead of calling schema_for_resource again.

christophebrun-forest (approved):
Good for me

PR #804, opened 2026-09-11
feat(collection): let a smart field declare the columns it needs, and go lazy where every consumer has

matthv:
Column half of the "let a smart field declare what it needs, and go lazy where every consumer has" ticket — the relation-preload half is a separate follow-up ticket, out of scope here. Stacked on `feature/optim-rbac-capabilities` (now including #797) — base branch is `feature/optim-rbac-capabilities`, not `main`.

## Why

A smart field's getter is `object.instance_eval(&block)` — it runs in the record's own context, so it can read any column, including transitively through model code the agent can't see. The agent can't infer what to select, so today it defensively loads every column the moment a single smart field is projected. This makes an opt-in declaration the only way to select less for a collection **carrying a smart field**.

That caveat matters: a collection with **no** smart field at all needs no opt-in and starts projecting immediately on this PR landing — a relation-less list that got `SELECT *` before now gets only what was requested. 6 existing specs needed a wider `fields:` for exactly this reason (they read a column off the AR object post-`perform` without requesting it). The `MissingAttributeValve` only wraps smart-field getters — an `after_find`, a `to_s` override, a `default_scope`, a decorator, or a smart action reading an unrequested column on one of these newly-projected collections is outside its reach, and would 500 instead of degrading.

## What

- A new `dependencies:` option on `field`/`has_many`/`belongs_to`, alongside the existing DSL options. A bare name is a column to select; an `a:b` path names a relation to preload (parsed and validated here, not yet consumed — that's the follow-up ticket). Absent, invalid, or normalized-away, a field is treated exactly as if it declared nothing.
- Adoption is per request, not per collection: a request only goes lazy-select when every smart field or smart relation it actually names is either a declared computed field or controlled some other way — `should_include_attr?` never evaluates a field the request didn't ask for, so one undeclared getter elsewhere on the collection can't affect a request that never names it. (An earlier version of this PR made it collection-wide instead — corrected after review: that version regressed every existing collection's `show`/`relationships` projection on day one, for a risk `should_include_attr?` already rules out.)
- `dependencies:` is accepted, normalized and validated on `has_many`/`belongs_to` exactly like on `field`, but nothing reads it back yet — requesting a smart relation, declared or not, still falls back to the unprojected path for that request. Wiring it up (so a declared smart relation's own columns can also narrow the select) is left to a follow-up.
- A safety valve for the case a declaration is incomplete rather than absent (a genuine mistake, or a getter reaching through model code the agent can't see): `MissingAttributeError` is rescued once, the record reloaded (its already-loaded associations preserved across the reload), and a warning logged naming the field and the missing column. A second failure degrades to `nil`, never a 500.
- The primary key and every foreign key the serializer will need to emit relationships are always selected regardless of the projection, the same guard the redaction ticket needed.
- Polymorphic associations are now batch-resolved through `ActiveRecord::Associations::Preloader` on Rails 6.1 too (previously gated to Rails ≥ 7, one query per record on 6.1) — including on the has-many relationship route, which didn't preload them at all before this.

## Known gaps

- A relation-path dependency's own preload is the follow-up ticket's job, not this one's — declaring one still makes its collection lazy-select-eligible without adding anything to the select for that relation. For a `has_many`/`has_one` path this costs nothing extra (the association loads via its own query regardless of the parent's select). For a `belongs_to` path, the association's own foreign key is proactively selected here specifically to avoid a wasted valve reload — but the target's own columns are still read via an ordinary lazy load, same as if no dependency had been declared at all.
- `SmartFieldDependencies::RelationPath`/`#relation_paths` are parsed and validated but have no consumer yet — dead code until the follow-up ticket lands, kept here so that ticket only adds the preloading, not a second parser.
- A polymorphic association reached only through a smart-field's own model code (not declared in `dependencies:`, not requested as a field) can still cost a query per record on the has-many relationship route under extended search — narrower than before this PR, not eliminated.

## Verification

839 examples, 0 failures, green on every supported Rails version (6.1 through 8.1). Before/after footprints, measured with the request-level query-count harness (`spec/requests/query_footprint_spec.rb`):

- **Polymorphic association on Rails 6.1**: 1+N queries (one per row, via each record's own lazy `belongs_to` load) → 1+1 (one batched query resolving every target), matching what Rails ≥ 7 already had.
- **A smart field walking a to-many relation** (`Owner#tree_names`, declaring only a relation-path dependency): root select narrows from `SELECT "owners".*` to the requested columns only; the per-row relation load itself (out of scope here) is unchanged.
- **A smart field reading an under-declared column**: correct value served, one extra query *per row* (the reload) — but the warning naming the field and column logs once per process, not once per row (fixed after review: it originally warned on every row, which would have meant thousands of identical lines on a large page). Never a 500 or a silently wrong value, pinned as the adoption-blocking criterion the ticket calls out.

fixes PRD-1086

> [!NOTE]
> ### Add smart-field dependency declarations and lazy projection to getters
> - Smart fields can now declare `dependencies` (bare columns or colon-delimited relation paths) so the server knows which columns to select. The new [smart_field_dependencies.rb](app/services/forest_liana/smart_field_dependencies.rb) normalizes and validates declarations at bootstrap, removing invalid ones with a warning.
> - [Collection](app/models/forest_liana/model/collection.rb) classifies computed smart fields (with declared dependencies) as projectable, while smart relations and integration fields are not. `ResourcesGetter`, `HasManyGetter`, and `ResourceGetter` use this check instead of rejecting projection whenever any smart field is requested.
> - Getters now save the unprojected relation before projecting and use it for counts, so projected column selection no longer breaks count queries.
> - A new [MissingAttributeValve](app/serializers/forest_liana/missing_attribute_valve.rb) serializer module catches `ActiveModel::MissingAttributeError`, reloads the record once (preserving loaded associations), retries evaluation, and degrades to a reported nil on failure. This guards cases where a declared dependency is incomplete or absent.
> - Polymorphic association preloading moves to a shared helper in [base_getter.rb](app/services/forest_liana/base_getter.rb), working on both Rails 6 and Rails 7 and assigning targets to the exact returned record objects.
> - Risk: smart-field getters that previously swallowed `MissingAttributeError` and returned nil now propagate it to the valve, which may reload the record and emit a process-level warning; collections with unresolved models are skipped during dependency validation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e5807e0.</sup>
> 
>

### DIFF
--- a/app/models/forest_liana/model/collection.rb
+++ b/app/models/forest_liana/model/collection.rb
@@ -70,4 +70,32 @@ class ForestLiana::Model::Collection
       .select { |field| field[:'is_virtual'] && field[:type] == 'String' }
       .map { |field| field[:field].to_s }
   end
+
+  # is_virtual alone also matches a smart relation (has_many/belongs_to, whose reference/
+  # integration is set) — dependencies only ever govern what a computed getter's own `select`
+  # needs, so those are excluded here.
+  def computed_smart_fields
+    fields.select { |field| field[:is_virtual] && field[:reference].nil? && field[:integration].nil? }
+  end
+
+  def smart_field_dependencies_declared?
+    computed_smart_fields.all? { |field| field.key?(:dependencies) }
+  end
+
+  def smart_fields_projectable?(field_names)
+    return false unless smart_field_dependencies_declared?
+
+    requested_names = field_names.map(&:to_s)
+    computed_smart_fields
+      .select { |field| requested_names.include?(field[:field].to_s) }
+      .all? { |field| field.key?(:dependencies) }
+  end
+
+  def smart_field_dependency_columns(field_names)
+    requested_names = field_names.map(&:to_s)
+    computed_smart_fields
+      .select { |field| requested_names.include?(field[:field].to_s) }
+      .flat_map { |field| ForestLiana::SmartFieldDependencies.for(field).columns }
+      .uniq
+  end
 end
--- a/app/models/forest_liana/model/collection.rb
+++ b/app/models/forest_liana/model/collection.rb
@@ -82,13 +82,11 @@ class ForestLiana::Model::Collection
     computed_smart_fields.all? { |field| field.key?(:dependencies) }
   end
 
-  def smart_fields_projectable?(field_names)
-    return false unless smart_field_dependencies_declared?
-
-    requested_names = field_names.map(&:to_s)
-    computed_smart_fields
-      .select { |field| requested_names.include?(field[:field].to_s) }
-      .all? { |field| field.key?(:dependencies) }
+  # All-or-nothing, deliberately: an undeclared computed smart field poisons projection for the
+  # whole collection even when the request never names it (smart_field_dependencies_declared?
+  # already checks every one of them) — there is no per-request subset to narrow to here.
+  def smart_fields_projectable?
+    smart_field_dependencies_declared?
   end
 
   def smart_field_dependency_columns(field_names)
@@ -96,6 +94,14 @@ class ForestLiana::Model::Collection
     computed_smart_fields
       .select { |field| requested_names.include?(field[:field].to_s) }
       .flat_map { |field| ForestLiana::SmartFieldDependencies.for(field).columns }
+      .uniq
+  end
+
+  def smart_field_dependency_relation_paths(field_names)
+    requested_names = field_names.map(&:to_s)
+    computed_smart_fields
+      .select { |field| requested_names.include?(field[:field].to_s) }
+      .flat_map { |field| ForestLiana::SmartFieldDependencies.for(field).relation_paths }
       .uniq
   end
 end
--- a/app/serializers/forest_liana/missing_attribute_valve.rb
+++ b/app/serializers/forest_liana/missing_attribute_valve.rb
@@ -10,25 +10,31 @@ module ForestLiana
       super
     rescue ActiveModel::MissingAttributeError => exception
       column = missing_column_from(exception)
-      raise unless reloadable?(object, column)
+      return degrade(attribute_name, exception) unless reloadable?(object, column)
 
-      reload_keeping_associations(object)
-
+      # #reload itself can raise (the row was deleted between the list query and serialization,
+      # a statement timeout, ...) — kept inside this same rescue so that degrades exactly like a
+      # second MissingAttributeError would, instead of escaping this valve entirely.
       begin
+        reload_keeping_associations(object)
         result = super
         FOREST_LOGGER.warn "Field \"#{attribute_name}\" of the \"#{type}\" collection read the " \
           "\"#{column}\" column without declaring it in dependencies: — reloaded the record to " \
           'serve it, at the cost of an extra query. Add it to the field\'s dependencies: to avoid this.'
         result
-      rescue ActiveModel::MissingAttributeError => second_exception
-        FOREST_REPORTER.report second_exception
-        FOREST_LOGGER.error "Cannot retrieve the \"#{attribute_name}\" value of the \"#{type}\" " \
-          "collection because of an internal error in the getter implementation: #{second_exception.message}"
-        nil
+      rescue ActiveModel::MissingAttributeError, ActiveRecord::ActiveRecordError => second_exception
+        degrade(attribute_name, second_exception)
       end
     end
 
     private
+
+    def degrade(attribute_name, exception)
+      FOREST_REPORTER.report exception
+      FOREST_LOGGER.error "Cannot retrieve the \"#{attribute_name}\" value of the \"#{type}\" " \
+        "collection because of an internal error in the getter implementation: #{exception.message}"
+      nil
+    end
 
     # A missing attribute already loaded on this exact record (as opposed to a relation's own
     # record, read through it) can never be fixed by reloading this record — degrade immediately
--- a/app/serializers/forest_liana/missing_attribute_valve.rb
+++ b/app/serializers/forest_liana/missing_attribute_valve.rb
@@ -1,0 +1,64 @@
+module ForestLiana
+  # Included into the serializer classes SerializerFactory#serializer_for generates (after
+  # ForestAdmin::JSONAPI::Serializer, which places it above that module in the MRO — `super` below
+  # still reaches the gem's own evaluate_attr_or_block). A smart field's dependencies: can be
+  # incomplete without ever being invalid (a genuine mistake, a getter that reads a column no one
+  # thought to declare) — this is the safety valve for that: one retry, never a 500 or a silently
+  # wrong value.
+  module MissingAttributeValve
+    def evaluate_attr_or_block(attribute_name, attr_or_block)
+      super
+    rescue ActiveModel::MissingAttributeError => exception
+      column = missing_column_from(exception)
+      raise unless reloadable?(object, column)
+
+      reload_keeping_associations(object)
+
+      begin
+        result = super
+        FOREST_LOGGER.warn "Field \"#{attribute_name}\" of the \"#{type}\" collection read the " \
+          "\"#{column}\" column without declaring it in dependencies: — reloaded the record to " \
+          'serve it, at the cost of an extra query. Add it to the field\'s dependencies: to avoid this.'
+        result
+      rescue ActiveModel::MissingAttributeError => second_exception
+        FOREST_REPORTER.report second_exception
+        FOREST_LOGGER.error "Cannot retrieve the \"#{attribute_name}\" value of the \"#{type}\" " \
+          "collection because of an internal error in the getter implementation: #{second_exception.message}"
+        nil
+      end
+    end
+
+    private
+
+    # A missing attribute already loaded on this exact record (as opposed to a relation's own
+    # record, read through it) can never be fixed by reloading this record — degrade immediately
+    # rather than pay for a query that will only fail the same way again.
+    def reloadable?(record, column)
+      column.present? && record.is_a?(ActiveRecord::Base) && record.persisted? &&
+        record.class.column_names.include?(column) && !record.has_attribute?(column)
+    end
+
+    # Rails 6.1 mutates @association_cache in place on #reload; Rails 8.1 replaces it with an
+    # empty one — either way, an already-eager-loaded relation would otherwise re-query at
+    # serialization time. Captured and restored around the reload so neither version's behavior
+    # costs this valve an extra join.
+    def reload_keeping_associations(record)
+      loaded_targets = record.class.reflect_on_all_associations.each_with_object({}) do |reflection, targets|
+        association = record.association(reflection.name)
+        targets[reflection.name] = association.target if association.loaded?
+      end
+
+      record.reload
+
+      loaded_targets.each { |name, target| record.association(name).target = target }
+    end
+
+    # NameError#name is nil for this exception on every Rails version this gem supports — the
+    # column has to come from the message instead. Format changed between Rails versions:
+    # "missing attribute: name" (<= 7.0) vs "missing attribute 'name' for Owner" (>= 7.1).
+    def missing_column_from(exception)
+      match = exception.message.match(/missing attribute:?\s+'?([a-zA-Z_][a-zA-Z0-9_]*)'?/)
+      match && match[1]
+    end
+  end
+end
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -71,21 +71,7 @@ module ForestLiana
       records = @records.offset(offset).limit(limit).to_a
       polymorphic_association, preload_loads = analyze_associations(@resource)
 
-      if polymorphic_association.any? && Rails::VERSION::MAJOR >= 7
-        preloader = ActiveRecord::Associations::Preloader.new(records: records, associations: polymorphic_association)
-        preloader.loaders
-        preloader.branches.each do |branch|
-          branch.loaders.each do |loader|
-            records_by_owner = loader.records_by_owner
-            records_by_owner.each do |record, association|
-              record_index =  records.find_index { |r| r.id == record.id }
-              records[record_index].define_singleton_method(branch.association) do
-                association.first
-              end
-            end
-          end
-        end
-      end
+      preload_polymorphic_associations(records, polymorphic_association)
 
       records
     end
--- a/app/services/forest_liana/resources_getter.rb
+++ b/app/services/forest_liana/resources_getter.rb
@@ -41,19 +41,22 @@ module ForestLiana
       # makes Rails emit `COUNT(col1, col2)`, invalid SQL, if `.count` ever ran off it instead.
       @unprojected_records = optimize_record_loading(@resource, @records, false)
 
-      has_smart_fields = Array(@params.dig(:fields, @collection_name)&.split(',')).any? do |field|
-        ForestLiana::SchemaHelper.is_smart_field?(@resource, field)
-      end
-
-      @records = if has_smart_fields
-        @unprojected_records
-      else
+      @records = if project?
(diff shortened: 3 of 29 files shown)


app/services/forest_liana/smart_field_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      FOREST_LOGGER.warn "Invalid dependency '#{invalid}' declared on smart field " \
+        "'#{field[:field]}' of the '#{collection_name}' collection: it does not resolve to a " \
+        'real column, or crosses a polymorphic relation. Ignored — the field is treated as if it ' \
+        'declared no dependencies at all.'
+      field.delete(:dependencies)
+    end
+
+    def self.valid_entry?(model, entry)
+      *relation_names, column_name = entry.split(':')
```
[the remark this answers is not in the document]
matthv: Confirmed and fixed — split(':', -1) preserves the trailing empty field so the entry is now correctly rejected instead of passing validate! as a bare column then crashing #relation_paths' caller. Regression test added (`fix(smart-field-dependencies)`, pushed).

app/services/forest_liana/smart_field_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+  # select, or an `a:b:c` relation path to preload — this class only splits the two apart; the
+  # preload side itself is unimplemented today, #relation_paths is ready for it as-is).
+  class SmartFieldDependencies
+    RelationPath = Struct.new(:relations, :column)
+
+    def self.normalize(raw)
+      return nil unless raw.is_a?(String) || raw.is_a?(Symbol) || raw.is_a?(Array)
+
+      entries = raw.is_a?(Array) ? raw : [raw]
+      entries.map(&:to_s).map(&:strip).reject(&:empty?).uniq
```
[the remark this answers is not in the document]
matthv: Confirmed and fixed — a nonempty input that normalizes down to nothing now answers nil (routed through the same "invalid, ignored" warning), distinct from an explicit `[]`. Regression test added (`fix(smart-field-dependencies)`, pushed).

app/services/forest_liana/resources_getter.rb
```diff
@@ -129,6 +122,11 @@ def includes_for_serialization
 
     private
 
+    # See Model::Collection#smart_fields_projectable? for why this is all-or-nothing per collection.
+    def project?
```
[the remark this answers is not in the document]
matthv: Confirmed — smart_fields_projectable? only gates on computed_smart_fields, so a block-backed smart relationship can read a column outside the select. Not a crash in the current state of this branch though: evaluate_attr_or_block is already wrapped by MissingAttributeValve (this same PR), which reloads once and degrades with a warning rather than raising. So the actual cost today is a silent per-record reload query on such collections, not a 500 — still real, and not yet called out in the PR's "Known gaps". Leaving this thread open pending a decision on whether to exclude those collections from projectability now or track it as a follow-up.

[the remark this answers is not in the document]
matthv: Fixed directly — a smart relation (belongs_to/has_many) or integration field now blocks projection for the whole collection (Model::Collection#smart_fields_projectable? checks a new uncontrolled_smart_fields, same all-or-nothing rule already applied to an undeclared computed field). Regression test added, mutation-checked, pushed.

app/models/forest_liana/model/collection.rb
```diff
@@ -70,4 +70,46 @@ def string_smart_fields_names
...
+  # without ever declaring dependencies:, so its mere presence has to block projection for the
+  # whole collection until it can declare and have those validated too (unimplemented today).
+  def uncontrolled_smart_fields
+    fields.select { |field| field[:is_virtual] && (field[:reference] || field[:integration]) }
+  end
+
+  # All-or-nothing, deliberately: an undeclared computed smart field poisons projection for the
+  # whole collection even when the request never names it (smart_field_dependencies_declared?
+  # already checks every one of them) — there is no per-request subset to narrow to here.
+  def smart_fields_projectable?
```
christophebrun-forest: **1. 🔴 The per-collection all-or-nothing rule is a measured regression on `show` and `/relationships/*`**

Before this PR, `project?` was **per request** (`projection? && !smart_field_requested?`): a collection carrying an undeclared smart field still projected as soon as the request didn't name it. After it, a single undeclared smart field turns projection off for **every** request on that collection.

Measured on the same fixture (`Location#alter_coordinates`, deliberately undeclared), same `Forest-Projection: id,coordinates` request:

```
base → SELECT "locations"."id", "locations"."coordinates" FROM "locations" WHERE id = ?
PR   → SELECT "locations".*                               FROM "locations" WHERE id = ?
```

On the day this ships, no customer has `dependencies:` anywhere. So for every collection carrying at least one smart field — most of them — this PR removes the projection they already had on `show` and on the has-many route, and only gives back the relation-less list. The net result is probably negative until adoption.

**What bothers me about the rationale.** Commit 3 says "simpler and safer than tracking which fields a given request will actually evaluate". But the PR itself states two lines below that `should_include_attr?` already gates evaluation on the projected fields — and that is accurate: when `project?` is true, `render_jsonapi` / `render_record_jsonapi` pass a non-nil `requested_fields` to the serializer, so `@_fields[type]` is populated and a smart field that wasn't requested is **never evaluated**. The risk the collection-wide rule claims to cover doesn't exist.

A per-request rule would be both safe and strictly better than the old *and* the new behavior:

```ruby
def project?
  projection? && @collection.requested_smart_fields_declared?(@field_names_requested)
end
# = computed_smart_fields.select { requested }.all? { |f| f.key?(:dependencies) }
```

If you'd rather keep the collection-wide rule, it should at least be stated explicitly in the description: it's a temporary regression until adoption, not a status quo.

app/models/forest_liana/model/collection.rb
```diff
@@ -70,4 +70,46 @@ def string_smart_fields_names
...
+  def smart_field_dependencies_declared?
+    computed_smart_fields.all? { |field| field.key?(:dependencies) }
+  end
+
+  # A smart relation (belongs_to/has_many) or integration field is is_virtual too but excluded
+  # from computed_smart_fields above — its own block can still read arbitrary root attributes
+  # without ever declaring dependencies:, so its mere presence has to block projection for the
+  # whole collection until it can declare and have those validated too (unimplemented today).
+  def uncontrolled_smart_fields
+    fields.select { |field| field[:is_virtual] && (field[:reference] || field[:integration]) }
```
christophebrun-forest, line 87: **2. 🔴 `dependencies:` on `has_many` / `belongs_to` is inert**

The description advertises "A new `dependencies:` option on `field`/`has_many`/`belongs_to`". On the latter two the option is accepted, normalized (`normalize_dependencies!` is indeed called) and validated at boot… yet it can never unlock anything, because `uncontrolled_smart_fields` doesn't look at `key?(:dependencies)`.

Verified by adding a **fully declared** smart belongs_to to the `Tree` fixture:

```
field hash  : {:reference=>"User.id", :dependencies=>["owner_id"], :field=>:smart_owner, :is_virtual=>true, ...}
uncontrolled: [:smart_owner]
projectable?: false        # despite a complete, valid declaration
```

A customer who follows the docs and diligently declares dependencies everywhere gets **no** projection at all as soon as they have one smart belongs_to — the common case. Two ways out: either `uncontrolled_smart_fields` excludes declared fields, or `dependencies:` isn't wired to `has_many`/`belongs_to` yet and the description shouldn't advertise it.

**Corollary on this very line**: the filter relies on `field[:reference]`, but neither `has_many` nor `belongs_to` in `lib/forest_liana/collection.rb` sets that key — only the user's own `opts` does. A smart relation declared without `reference:` therefore falls back into `computed_smart_fields` and its declared columns end up in the `SELECT` — the opposite classification for the same DSL call.

The unit spec builds the hash by hand (`{ reference: 'Other.id' }`) so it can't catch the discrepancy; and no `spec/dummy` fixture contains a smart relation at all, so `uncontrolled_smart_fields` has no request-level coverage.

app/serializers/forest_liana/missing_attribute_valve.rb
```diff
@@ -0,0 +1,70 @@
...
+      column = missing_column_from(exception)
+      return degrade(attribute_name, exception) unless reloadable?(object, column)
+
+      # #reload itself can raise (the row was deleted between the list query and serialization,
+      # a statement timeout, ...) — kept inside this same rescue so that degrades exactly like a
+      # second MissingAttributeError would, instead of escaping this valve entirely.
+      begin
+        reload_keeping_associations(object)
+        result = super
+        FOREST_LOGGER.warn "Field \"#{attribute_name}\" of the \"#{type}\" collection read the " \
```
christophebrun-forest: **3. 🟠 The valve's cost is per record, not per request**

The description says "one extra query" and "one warning". In practice `evaluate_attr_or_block` runs once per (record × field): a page of 100 rows with one under-declared field means **100 reloads and 100 log lines** — strictly worse than the `SELECT *` this PR replaces.

The spec (`expect(FOREST_LOGGER).to receive(:warn).once`) only covers a single record, which hides the behavior entirely at volume.

The extra query is inherent to the design and acceptable. The logging isn't: warning once per `(collection, field)` per process would get the adoption message across without drowning production logs. Failing that, at least fix the wording in the description.

app/services/forest_liana/smart_field_dependencies.rb
```diff
@@ -0,0 +1,77 @@
...
+      @entries = entries
+    end
+
+    def columns
+      @entries.reject { |entry| entry.include?(':') }
+    end
+
+    def relation_paths
+      @entries.select { |entry| entry.include?(':') }.map do |entry|
+        *relations, column = entry.split(':')
```
christophebrun-forest: **4. 🟠 Possible `NoMethodError` (500) — the `split(':')` / `split(':', -1)` asymmetry is only half fixed**

Commit 33c4d24 fixed `valid_entry?` to use `split(':', -1)`, but this line still uses `split(':')`. Verified:

```
relation_paths for 'name:' → [RelationPath relations=[], column="name"]
relations.first.to_sym     → NoMethodError: undefined method `to_sym' for nil
normalize('name:')         → ["name:"]      # normalize does not reject it
```

So the only guard is `validate!` at boot — but `generate_apimap` is gated on `if ForestLiana.env_secret` (bootstrapper.rb:41) and `validate_smart_field_dependencies` does `next if model.nil?`. Outside that path, the entry reaches `compute_select_fields`, where `get_one_association(relation_path.relations.first)` blows up outside every `rescue` → a 500 on every list of that collection.

One-line defense in depth, either:

```ruby
next if relation_path.relations.empty?      # in compute_select_fields
```
or align this line on `split(':', -1)` the way `valid_entry?` does.

christophebrun-forest (commented):
Detailed review. Full suite re-run locally on Rails 6.1 (836 examples, 0 failures) and CI green on all 6 gemfiles — nothing below is a broken test. Three of the points were verified experimentally (throwaway probes); the measurements are in the comments.

The work is careful and the infrastructure (parser, boot-time validation, valve, shared polymorphic preloader) is the right one. Two points look blocking to me before merge (1 and 2): taken together, they make this PR *remove* projection from nearly every existing collection, and the adoption path it documents unlocks nothing as soon as a smart belongs_to is present. Points 4 and 5 are one-line fixes. Points 3 and 6 are mostly about the description / release notes.

app/services/forest_liana/resources_getter.rb
```diff
@@ -36,24 +36,31 @@ def assert_sort_readable!
...
-      has_smart_fields = Array(@params.dig(:fields, @collection_name)&.split(',')).any? do |field|
-        ForestLiana::SchemaHelper.is_smart_field?(@resource, field)
-      end
-
-      if includes.empty? || has_smart_fields
-        @records = optimize_record_loading(@resource, @records, false)
+      # Captured before any select narrows @records: count/optimized_count must build its COUNT
+      # off this, never off @records past this point — a `.select` naming more than one column
+      # makes Rails emit `COUNT(col1, col2)`, invalid SQL, if `.count` ever ran off it instead.
+      @unprojected_records = optimize_record_loading(@resource, @records, false)
```
christophebrun-forest: **5. 🟠 `@unprojected_records` isn't idempotent**

Here and in `HasManyGetter#perform` (line 35), `@unprojected_records` is reassigned on every `perform`. A second `perform` on the same instance captures it **after** projection — exactly the `COUNT(col1, col2)` the guard exists to prevent.

No current caller does this, so it's latent, but `@unprojected_records ||=` costs two characters and makes the guard genuinely unconditional.

app/services/forest_liana/resources_getter.rb
christophebrun-forest, line 127: **6. 🟡 Projecting relation-less lists is a behavior change with no opt-in**

The description says "without one, behavior is unchanged". That holds for collections *with* smart fields, not for those without any: `computed_smart_fields.all?` on an empty list is `true`, so every smart-field-free collection becomes projectable — including relation-less lists, which used to get `SELECT *`.

The clearest signal is in the PR itself: **6 existing specs** had to widen their `fields:` (`'Tree' => 'id'` → `'id,name'`, `'Island' => 'id'` → `'id,name'`, …) because they read a column off the AR object after `perform`. That is exactly the failure mode customers will hit — except on their side it will be an `after_find`, a `to_s`, a `default_scope`, a decorator or a smart action, all **out of the valve's reach** (it only wraps `evaluate_attr_or_block`). There it's a 500, not a degradation.

Worth stating in the description and the release notes, and possibly putting behind a flag for a first step.

app/models/forest_liana/model/collection.rb
matthv: Confirmed and fixed — smart_fields_projectable? now takes the request's own field names: should_include_attr? never evaluates a field the request didn't name, so an undeclared field elsewhere on the collection can't affect a request that never names it either. The description was wrong to describe this as a per-collection tradeoff — corrected, and noted there as a regression fixed after review.

app/models/forest_liana/model/collection.rb
matthv: Confirmed on both points. The all-or-nothing-per-collection fix above already scopes this to a per-request check, so a smart relation only blocks the specific request that names it, not the whole collection. But dependencies: on has_many/belongs_to is still genuinely inert (uncontrolled_smart_fields never reads it back) — corrected the description rather than wiring it up now, since narrowing a relation's own select needs more than this PR's scope. Left the field[:reference] corollary as-is: it's the same convention fields_smart_belongs_to already relies on (a smart relation is expected to always pass reference:), not something new to this PR.

app/serializers/forest_liana/missing_attribute_valve.rb
matthv: Confirmed and fixed — the warning now dedupes per (collection, field, column) for the life of the process instead of firing on every record. Description corrected too (it undersold the per-row query cost as "one extra query").

app/services/forest_liana/smart_field_dependencies.rb
matthv: Confirmed and fixed — #relation_paths now also uses split(':', -1), matching valid_entry?. Regression test added.

app/services/forest_liana/resources_getter.rb
matthv: Fixed — ||= on both. Cheap insurance, agreed.

app/services/forest_liana/resources_getter.rb
matthv: Confirmed — this is intended (a smart-field-free collection needs no opt-in), but the description undersold the risk. Added an explicit callout: the valve only covers smart-field getters, so an after_find/to_s/default_scope/decorator/smart action reading an unrequested column on one of these collections is outside its reach and would 500. Not gating it behind a flag for now — flagging it prominently in the description is the tradeoff for this round.

christophebrun-forest (approved):
LGTM

PR #805, opened 2026-09-15
perf(getters): preload the relations a smart field declares, instead of walking them per row

christophebrun-forest:
Closes [PRD-1089](https://linear.app/forestadmin/issue/PRD-1089).

Targets the shared branch `feature/optim-rbac-capabilities`, where #804 (the `dependencies:` option and its column half) and #802 have both landed. Nothing to read first any more.

## Why

A smart field getter runs as `object.instance_eval(&block)`, **once per record**. A getter that reaches an association is therefore one or more queries **per listed row**, and nothing in the agent could see it coming.

#804 introduced `dependencies:` and implemented its **column** half — what to `select`. A path entry (`orders:amount`, `account:owner:email`) was parsed and validated, and then ignored. This PR gives it its behaviour: the relation it names is preloaded once for the whole page.

One declaration, both halves — no second vocabulary, no second parser.

## Before / after

Measured on a 25-row page, with the same subscriber the footprint harness uses:

| declaration | before | after |
|---|---|---|
| `trees:name` (to-many) | **26** queries — 25 `SELECT`s on `trees`, one per row | **2** — one `SELECT … WHERE owner_id IN (…)` |
| `island:location:coordinates` (two hops) | **51** queries — 25 on `isle` + 25 on `locations` | **3** — one per hop |

The specs assert row-count invariance rather than those absolute numbers, per PRD-1093: the same request is run against 2 rows and against 10, and the counts must be equal.

## What changed

- `BaseGetter#smart_field_preloads` turns the relation paths of the smart fields **actually being serialised** into one nested `preload` argument (`['account','owner']` → `{account: {owner: {}}}`, paths sharing a first hop deep-merged into one branch).
- Wired into `ResourcesGetter#perform` (list) and `HasManyGetter#perform` (related list), in both cases **after** the capture that guards `count` — a `COUNT` has no getter to run and must not inherit a preload.
- `BaseGetter#preload_owner_keys` selects the column the Preloader reads off the owner row, so a narrowed select still carries it. That column is `join_foreign_key`, walked down to the first hop of a `:through` — see Safety, it is not "the belongs_to's foreign key" in three separate shapes.
- `get_record` carries the association cache across `becomes()`, in both controllers: `init_internals` has just emptied it, so an STI collection dropped every preload and re-queried once per row. This also restores the relations eager loaded for serialization, which `becomes()` was already discarding.
- Get-one is deliberately left out: with a single record a preload costs exactly what the lazy load costs.

**Only the serialised fields contribute.** A collection carrying several declared smart fields where the request projects one must not preload the others' relations, or this trades an N+1 for a constant over-fetch and the win evaporates. Pinned by its own scenario.

## Two deliberate deviations, both worth arguing with

**1. The preload loads the target's full rows — it does not narrow to the declared column.**

The ticket describes a path as "an association to preload, *and a column to select on the target*". I did the first and not the second, on purpose: `MissingAttributeValve` can only repair a missing attribute on the **root** record — that is exactly what #804's `Tree#owner_name` fixture pins. Narrowing a preloaded target's columns would turn a merely incomplete declaration into a silently wrong value with nothing behind it to catch it. The `scope:` keyword on `Preloader` is also not uniform across 6.1 → 8.1.

**2. "A declared relation that does not exist fails at declaration time naming the path."**

It is named at boot, but nothing fails: #804 chose warn-and-ignore in `SmartFieldDependencies.validate!` so that a bad declaration cannot stop an existing agent from booting. Inherited here rather than diverged from. Flagging it so it stays a decision.

## Safety

A path is skipped — keeping today's lazy load, never raising — when it names a relation that does not resolve, crosses a polymorphic one (which cannot be joined and which boot validation already rejects), or is instance-dependent on Rails 6.1, whose `Preloader` refuses those outright (the condition mirrors its own `check_preloadable!`, so an optional or splat argument — arity `-1` — fails it too). Same version gate `optimize_record_loading` already applies to its own preload.

That check walks the **whole** chain, not only the hops the path names: `Preloader::ThroughAssociation` re-enters the preloader on the through and source reflections, so an intermediate hop's scope reaches `check_preloadable!` too. Its raise is an `ArgumentError` at query-resolution time — neither `NameError` nor `ActiveRecordError`, so no rescue here reaches it, and the whole list goes rather than the one field. Reachable on the dummy app as it stands (`Island#eponymous_tree` carries such a scope).

**Every skip is logged, once per (collection, path, reason).** Each one silently reinstates the N+1 the declaration was written to remove, and what breaks a working declaration is usually a change made elsewhere long after anyone verified it — a scope added to an association a path happens to cross. The line names the hop at fault, which is not always the hop declared. Same `warn_once` shape `MissingAttributeValve` already uses next door.

**The owner-side key has to be selected, or the preload breaks the list out of reach of every valve.** `MissingAttributeValve` runs at serialization time; this happens *resolving the query*, so it never sees it. `join_foreign_key` is that key in every shape, and it is not the belongs_to FK in three of them, each reproduced before being fixed and each pinned by its own scenario:

| first hop | key actually read | was |
|---|---|---|
| `belongs_to` to a model excluded from the schema | `trees.island_id` | dropped by `get_one_associations`' `model_included?` filter |
| `has_many … primary_key: 'name'` | `owners.name` | `owners.id` |
| `has_one … through: :island` | `trees.island_id` | `trees.id` |

How that failure presents depends on the version, which is worth stating because the quieter half is the worse half. Neutralising this selection, the two `:through` scenarios 500 with `ActiveModel::MissingAttributeError` on every version (`BelongsToAssociation#stale_state` passes a raising block to `_read_attribute`). The two `has_many … primary_key:` scenarios 500 on Rails 6.1 — but on 7.1/8.1 they answer **HTTP 200 with a blank value**, no error and no log, because `Preloader::Association#derive_key` reads the key with no raising block there. So on modern Rails the same author mistake degrades to silently wrong data rather than to a visible failure. The specs assert on values and query counts, so they catch it either way.

**Known gap, not fixed here.** An association whose own scope narrows its `select` and drops its linking key (`has_many :x, -> { select(:name) }`) raises the same error, on the *target* row this time. It works lazily — a `WHERE`, no remapping — and breaks under `preload`, so this PR is what turns it. Catching it means evaluating the scope, a detection mechanism of its own rather than one more rule in `skip_preload?`; tracked separately rather than smuggled in here.

**Nothing here is opt-out.** A getter with no declaration keeps the behaviour and the query count it has today, N+1 included, pinned by its own scenario on a collection that declares nothing.

## Verification

- `spec/requests/query_footprint_spec.rb`: the to-many scenario's pinned N+1 flipped from `per_row_delta 1` to `0` — the win is in the spec diff, not in this description. Six scenarios added, 11 examples: multi-hop chain (and the isolation case, where the declared field the request never names stays unloaded), undeclared-keeps-its-N+1, related list through `HasManyGetter`, a first hop pointing at an excluded model, one keyed on `primary_key:`, and one reached through a `:through`.
- `spec/services/forest_liana/base_getter_spec.rb`: 9 unit specs on the preload tree itself — nesting, merging, each skip rule.
- `spec/services/forest_liana/base_getter_spec.rb` also carries 3 specs on the skip log itself — what it names, that it says it once per process rather than once per page, and that a path it can preload leaves nothing behind.
- `spec/requests/sti_record_conversion_spec.rb`: 4 specs on what survives `becomes()` — the association cache on the list and on the related list, a preloaded polymorphic target (which rides in that same cache, which is why `preload_polymorphic_associations` calls `#call` and not `#loaders` on 7+), and an entry the two classes do **not** reflect identically, which must *not* be carried. No dummy collection is STI (`Town < Location`, but `locations` carries no type column), so the conversion is driven directly rather than through a schema change whose only purpose would be to reach it; `Town` gains a deliberately divergent `island` for the last of those.
- **The guards are proven, not assumed:** neutralising `smart_field_preloads` fails 6 of the new scenarios, the related-list one included; neutralising the owner-key selection fails 4 (see the version note under Safety for how each presents); neutralising the `:through` walk, the reflection check on the cache copy, or `#call` fails one new spec each.
- 880 examples pass locally on Rails 6.1, 7.0, 7.1, 7.2, 8.0 and 8.1.

## Not in scope

The CSV export walks `query_for_batch`, untouched here, so it keeps its N+1 — volume-wise that is the worst case and it deserves its own change rather than a quiet rider on this one.

> [!NOTE]
> ### Preload declared smart-field relations in `BaseGetter` instead of loading per row
> - Adds a preload planner in [base_getter.rb](https://github.com/ForestAdmin/forest-rails/pull/805/files#diff-b8f760393646995fca05cc29342760c53c6e7bb3a6544cedc7d351b2b8ebf315) that converts declared smart-field relation paths into nested Active Record preload trees, merging shared prefixes and skipping polymorphic, missing, or instance-scope associations via `skip_preload?`.
> - Updates `compute_select_fields` to include the owner-side keys required for preloading custom-key and through associations, and adds `preload_owner_keys` to resolve the actual join key through reflections.
> - Applies the preload step in [resources_getter.rb](https://github.com/ForestAdmin/forest-rails/pull/805/files#diff-de9538c4fb8386da065cc3f2d8e37030e67a693d8261634ed0153ded9ef60782) and [has_many_getter.rb](https://github.com/ForestAdmin/forest-rails/pull/805/files#diff-fb435be68d698be6e543c28cdb467f043e3c972ffe5164e459998ab5c206b857) after record selection, on both projected and unprojected queries.
> - Preserves the association cache when converting STI records in [resources_controller.rb](https://github.com/ForestAdmin/forest-rails/pull/805/files#diff-be3235120644036caa6865b266de1f6d79eb2d247f996aeae24397e241822c27) and [associations_controller.rb](https://github.com/ForestAdmin/forest-rails/pull/805/files#diff-d87aaccd92079688766587630744649f7ba2d87e3a0864816d740b161c527054) so preloaded associations survive the collection-class switch.
> - Behavioral Change: list queries for resources and related records now issue one batch preload per declared smart-field relation hop instead of one query per row; queries also select additional key columns (`preload_owner_keys`) needed by the preloader. Smart fields that access associations without declaring a dependency still lazy-load per row.
>
> 
> #### Changes since #805 opened
>
> - Implemented preload-skip detection and warning system in `forest_liana.BaseGetter` to identify and log unpre-loadable association paths [2d37749]
> - Fixed `forest_liana.BaseGetter.preload_polymorphic_associations` on Rails >= 7 to write polymorphic association targets into the association cache [2d37749]
> - Introduced `forest_liana.ApplicationController.carry_preloaded_associations` helper to safely preserve compatible preloaded associations across STI `becomes()` calls [2d37749]
> - Updated `forest_liana.AssociationsController.get_record` and `forest_liana.ResourcesController.get_record` to use `carry_preloaded_associations` for STI conversion [2d37749]
> - Added test coverage for preload skipping logic, polymorphic association preservation, and STI conversion behavior [2d37749]
> - Added test case verifying `ForestLiana::BaseGetter#preload_polymorphic_associations` degradation behavior when reflection chain resolution fails [f45299f]
> - Updated inline documentation in `ForestLiana::BaseGetter` rescue block for reflection resolution failures [f45299f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a7f255.</sup>
> 
>


app/services/forest_liana/base_getter.rb
```diff
@@ -41,6 +41,60 @@ def optimize_record_loading(resource, records, force_preload = true)
       result
     end
 
+    # The relation half of a smart field's dependencies:. Its column half reaches the select
+    # (compute_select_fields); a path instead names an association the getter walks inside
+    # instance_eval, once per record — loaded here in one query for the whole page instead.
+    def apply_smart_field_preloads(records)
+      preloads = smart_field_preloads
+
+      preloads.empty? ? records : records.preload(preloads)
```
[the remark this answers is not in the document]
christophebrun-forest: Confirmed and fixed in 7c75289 — though the trigger is not the scope, it is the owner foreign key.

Reproduced with `ForestLiana.excluded_models = ["Island"]` and a `island:location:coordinates` dependency: HTTP 500, `missing attribute: island_id`. The FK was being selected through `QueryHelper.get_one_associations`, which filters on `model_included?` and therefore drops an association whose target is not an exposed collection — leaving exactly that relation preloaded with no key to preload it by.

It now reads the raw reflection, and is driven off `serialized_smart_field_names`, the same field set the preload uses, so the select and the preload cannot drift apart again. Pinned by a request spec that fails with the 500 without the fix.

app/services/forest_liana/base_getter.rb
```diff
@@ -41,6 +41,60 @@ def optimize_record_loading(resource, records, force_preload = true)
...
+    def skip_preload?(path)
+      model = projected_resource
+
+      path.relations.any? do |name|
+        association = model.reflect_on_association(name.to_sym)
+        next true if association.nil? || SchemaUtils.polymorphic?(association)
+        # Rails 6.1's Preloader refuses an instance-dependent scope outright; optimize_record_loading
+        # gates its own preload on the same version for the same reason. Falling back to the lazy
+        # load keeps today's N+1 there — slower than it could be, never wrong.
+        next true if Rails::VERSION::MAJOR < 7 && association.scope&.arity.to_i.positive?
```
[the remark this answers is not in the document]
christophebrun-forest: Correct, fixed in 7c75289. `check_preloadable!` raises `unless scope.arity == 0`, so `-1` — an optional or splat argument — fails it just as surely as `1`, and `positive?` waved it straight into the `ArgumentError` the rule exists to avoid.

Took the condition from Rails rather than `nonzero?`, so it reads as the mirror of `check_preloadable!` it is:

```ruby
next true if Rails::VERSION::MAJOR < 7 && association.scope && !association.scope.arity.zero?
```

Unit spec added with a stubbed splat scope.

app/services/forest_liana/resources_getter.rb
```diff
@@ -49,7 +49,10 @@ def perform
         @unprojected_records
       end
 
-      @records
+      # Deliberately after the capture above: a preload is a query of its own, per page — count
+      # builds its COUNT off @unprojected_records and has no getter to run, so it must not inherit
+      # one.
+      @records = apply_smart_field_preloads(@records)
```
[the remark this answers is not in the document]
christophebrun-forest: Real, and fixed in 7c75289 — in the controller rather than here, since that is where it breaks.

Verified on Rails 6.1 and 8.1: `becomes` copies `@attributes` onto a freshly allocated instance, but `init_internals` has just emptied `@association_cache` and nothing copies it back. `association(:island).loaded?` goes `true` -> `false` across the call. So an STI collection dropped the whole preload and re-queried per row.

`get_record` now carries the cache over, in both `ResourcesController` and `AssociationsController`. Worth noting this was already discarding the relations eager loaded for serialization, and the singleton readers `preload_polymorphic_associations` defines — so STI collections were paying for preloading twice over before this PR.

No dummy collection is STI (`Town < Location`, but `locations` has no type column), so the conversion is pinned directly rather than with a schema change whose only purpose would be to reach it.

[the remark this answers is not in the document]
christophebrun-forest: Fair, and deliberate — flagging the reasoning rather than the fix.

The waste needs all three of: no `fields[]` at all, a role that cannot read a declared smart field, and that field declaring a to-many. The cost is then bounded — one query per declared path, independent of page size — not an N+1.

The alternative is to drop the no-projection fallback entirely, one line, and simply not preload when the request projects nothing. I kept it because without `fields[]` every smart field *is* serialised and every getter *does* run, so that is precisely the case with the worst N+1 today; a role-redacted field is the exception there, not the rule, and paying a bounded over-fetch in the exception to fix the N+1 in the normal case seemed the better trade.

Filtering by what survives `redact_fields` is not available where the decision is made: that runs in the controller, after the getter has already built its query. Happy to drop the fallback instead if you would rather not carry the trade — say the word.

app/services/forest_liana/base_getter.rb
```diff
@@ -41,6 +41,62 @@ def optimize_record_loading(resource, records, force_preload = true)
...
+
+    # SmartFieldDependencies.validate! already rejects at boot a path that resolves to nothing or
+    # crosses a polymorphic relation — re-checked here so that a collection reaching this outside
+    # that pass degrades to the lazy load it has always done, rather than raising once per request.
+    def skip_preload?(path)
+      model = projected_resource
+
+      path.relations.any? do |name|
+        association = model.reflect_on_association(name.to_sym)
+        next true if association.nil? || SchemaUtils.polymorphic?(association)
```
matthv, line 85: 🔴 **`skip_preload?` degrades silently and permanently — no log, no report, ever**

Every branch here returns `true` (fall back to the pre-existing lazy N+1) with zero observability: no `FOREST_LOGGER` call anywhere in this method, unlike `MissingAttributeValve#warn_once` next door, which logs once per process for a conceptually similar "declared dependency isn't quite working" situation.

Concretely: a customer declares `account:owner:email`, it preloads correctly, it's verified once. Later someone adds an instance-dependent scope to `Account#owner` for an unrelated reason on a Rails 6.1 app. From that deploy on, this method returns `true` on every single request, forever — the optimization silently regresses back to full N+1 and nothing in the system ever says so. Same for the `rescue NameError, ActiveRecord::ActiveRecordError` at the bottom: `ActiveRecordError` is the base class of nearly every AR exception, and anything caught there degrades identically to "this relation doesn't exist," with nothing logged.

Worth a `warn_once`-style log (keyed on collection/field/path) on every branch that returns `true`, and/or narrowing the rescue to `NameError` alone.

app/services/forest_liana/base_getter.rb
```diff
@@ -41,6 +41,62 @@ def optimize_record_loading(resource, records, force_preload = true)
...
+        # Rails 6.1's Preloader refuses an instance-dependent scope outright; optimize_record_loading
+        # gates its own preload on the same version for the same reason. Falling back to the lazy
+        # load keeps today's N+1 there — slower than it could be, never wrong. Mirrors
+        # check_preloadable!'s own `scope.arity == 0`, which a scope taking an optional or splat
+        # argument (arity -1) fails just as surely as one taking a required one.
+        next true if Rails::VERSION::MAJOR < 7 && association.scope && !association.scope.arity.zero?
```
matthv, line 91: 🔴 **This guard doesn't walk a `:through` chain, so an intermediate hop's scope isn't checked — and the fallback `rescue` doesn't catch what that raises**

The Rails-6.1 arity check only inspects `reflect_on_association(name).scope` for each *declared* hop. But `Preloader::ThroughAssociation` (6.1) re-enters the preloader for both `through_reflection.name` and `source_reflection.name`, so an intermediate hop's scope is checked by `check_preloadable!` too — this method never looks at it.

So a path like `dependencies: ['x:col']` where `x` is `has_many … through: :y` and `y` carries an instance-dependent scope passes this guard, then raises `ArgumentError` at query resolution — which is neither `NameError` nor `ActiveRecordError`, so the method-level `rescue` doesn't catch it either. That takes down the whole list, not the one field.

This is reachable today, not theoretical: the dummy app already has such a scope (`Island#eponymous_tree`). Worth walking `through_reflection`/`source_reflection` in the arity check the same way `preload_owner_keys` below already walks it for the key.

app/controllers/forest_liana/resources_controller.rb
```diff
@@ -241,8 +241,17 @@ def is_sti_model?
...
+    #         association cache, which init_internals has just emptied. Everything the getters
+    #         preloaded (smart_field_preloads, and the relations eager loaded for serialization)
+    #         would be dropped here and re-queried once per row, so an STI collection alone would
+    #         keep the N+1 the rest of this pipeline exists to remove. Carried over explicitly.
     def get_record record
-      is_sti_model? ? record.becomes(@resource) : record
+      return record unless is_sti_model?
+
+      record.becomes(@resource).tap do |became|
+        became.instance_variable_set(:@association_cache, record.instance_variable_get(:@association_cache))
```
matthv, line 253: 🟠 **The copied association cache isn't checked against the target class's own reflections — could silently serve a different association's data under the base class's field name**

`@association_cache` holds live association proxies bound to the reflection of the class that built them (`record`'s actual runtime class). `association(name)` returns whatever's cached under that key without ever re-checking it against the *new* class's own reflection for that name. If a subclass overrides the same-named association (a common STI pattern — different `class_name:`, different scope, different `foreign_key`), the copied cache silently carries the *subclass's* semantics onto an instance now typed and serialized as the *base* class. No exception either way: if the two target classes share attribute names it serializes plausible-but-wrong data; if not, it's more likely a `NoMethodError` downstream — but nothing here checks.

Same in `associations_controller.rb#get_record`. Not exercised by `sti_record_conversion_spec.rb` either: that spec converts `Tree` to `Tree` itself (the dummy app has no real STI model with divergent associations), so the fix is only validated for the no-op-class-switch case — the only case where this failure mode is actually reachable is untested.

matthv, line 253: 🟠 **Polymorphic preloads are dropped here, not carried over — the comment overstates what this covers**

`assign_preloaded_targets` attaches a preloaded polymorphic target via `record.define_singleton_method(association_name)`, not through `@association_cache`. `becomes()` allocates a *new* object with an empty singleton class, so that attachment is lost — only the `@association_cache` copy happens here.

Verified directly: `singleton_before=[:addressable]` → `singleton_after=[]`, with a query re-issued after `becomes()`. So an STI collection carrying a polymorphic to-one keeps exactly the N+1 this PR exists to remove — the fix is incomplete for that one shape, and the comment above this method ("Everything the getters preloaded … Carried over explicitly") reads as complete coverage when it isn't. Either call `get_record` before `preload_polymorphic_associations` runs, re-apply `assign_preloaded_targets` after `becomes()`, or narrow the comment's claim.

app/services/forest_liana/base_getter.rb
```diff
@@ -328,6 +392,23 @@ def select_foreign_keys(select, owner, association, joined = true)
...
+    # column of the owner table at all: what the preloader reads there is the key of the hop it
+    # starts with, so the chain is walked down to that first hop before asking.
+    #
+    # Getting this wrong does not cost the one field — it raises resolving the query, where
+    # MissingAttributeValve (a serialization-time valve) never sees it, and takes down the list.
+    def preload_owner_keys(association)
+      reflection = association
+      reflection = reflection.through_reflection while reflection.through_reflection?
+
+      Array(reflection.join_foreign_key)
```
matthv, line 409: 🟡 **The "4 failures, each on a 500" claim only holds on Rails 6.1 — on 7.x/8.x, 2 of the 4 fail silently instead**

Mutation-checked (removing the owner-key selection this method feeds): on Rails 6.1 all 4 pinned scenarios die with a real `missing attribute` 500, as the PR body's Safety table states. On Rails 7.1/8.1 only the two `:through` scenarios still 500 (that path goes through `BelongsToAssociation#stale_state`, which passes a raising block to `_read_attribute`). The two `has_many`-with-custom-`primary_key:` scenarios instead return HTTP 200 with blank smart-field values — no error, no log — because plain `has_many` preloading (`Preloader::Association#derive_key`) reads the key with no raising block, so a missing attribute quietly evaluates to `nil`.

Doesn't affect the specs' own pass/fail (they assert on values/query counts, which still catch the regression on every version) — but worth a one-line correction in the PR description, since "each on the missing-attribute 500" undersells that the same author mistake degrades to silently wrong data on modern Rails instead.

app/services/forest_liana/base_getter.rb
christophebrun-forest: Confirmed and fixed in 2d37749. Reproduced exactly as described, on Rails 6.1:

```
refl through? true  scope=nil  through_scope_arity=1
skip_preload? = false
preload RAISED ArgumentError: The association scope 'eponymous_tree' is instance dependent…
```

And `ArgumentError` is neither `NameError` nor `ActiveRecordError`, so the method-level rescue was never going to see it — it is raised resolving the query, not inside `skip_preload?`, so no rescue placed there could have.

The check now walks `through_reflection` / `source_reflection` recursively, the same descent `preload_owner_keys` already makes for the key. Two specs: the tree it produces, and — the one that actually reproduces the bug — that `apply_smart_field_preloads(...).to_a` resolves rather than raising. Both fail if the walk is neutralised. The log line below names the hop at fault, which is the intermediate one, not the declared one:

```
The "location:coordinates" dependency of the "Tree" collection cannot be preloaded
("island" has an instance-dependent scope, which Rails 6.1.7.9 cannot preload) — …
```

app/services/forest_liana/base_getter.rb
christophebrun-forest: Fair, and fixed in 2d37749 — `skip_preload?` now logs on every branch that returns `true`, once per `(collection, path, reason)` for the life of the process, the same `warn_once` shape `MissingAttributeValve` uses next door. The reason string names the hop at fault rather than the hop declared, which for a `:through` is not the same thing.

On narrowing the rescue: I kept `ActiveRecordError` alongside `NameError` rather than dropping it. `through_reflection.klass` raises `HasManyThroughAssociationNotFoundError` for a chain naming a hop that no longer exists — an `ActiveRecordError`, and genuinely the same "does not resolve" the first branch answers for, just reached one level down. What made it indistinguishable was that nothing was logged; now the class and message go into the line, so an unexpected one is visible rather than inferred. Commented to that effect.

Three specs on the log itself: what it names, that it says it once and not once per page, and that a path it can preload leaves nothing behind.

app/controllers/forest_liana/resources_controller.rb
christophebrun-forest: Real, and fixed in 2d37749. Reproduced with an STI pair whose subclass redefines the same association with a different foreign key:

```
sub_owner                    = cutter2
base_owner_with_copied_cache = cutter2   ← the subclass's semantics, under the base class's field
base_owner_fresh             = owner1
```

Plausible and wrong, nothing raised, exactly as you said.

`carry_preloaded_associations` (shared now by both controllers, since getting this wrong is a correctness bug rather than a difference of route) copies only the entries both classes reflect identically — `record.class.reflect_on_association(name) == became.class.reflect_on_association(name)`. Reflection equality compares `active_record`, so a subclass's override never matches its parent's even when the options are the same; the entry is then simply left out and loads lazily, the behaviour it had before the copy existed. Erring toward the extra query rather than the wrong value.

On the test gap: you were right that the only reachable case was untested. `Town` (which is already `< Location` in the dummy app) now carries a deliberately divergent `island`, and `sti_record_conversion_spec.rb` pins that the entry is *not* carried and that the base class reads back its own target. It fails without the reflection check.

christophebrun-forest: Half right, and the half that is right is worse than described — fixed in 2d37749, though not where you suggested.

The singleton readers are indeed lost: `singleton_before=[:addressable]` → `singleton_after=[]`, confirmed. But the targets do not depend on them — Rails' own Preloader writes a polymorphic target into `@association_cache` as well, so on Rails 6.1 they were riding along with the cache copy the whole time and no query was re-issued.

On Rails 7+ they were not, and that is the actual bug. `preload_polymorphic_associations` called `preloader.loaders`, not `preloader.call`. `records_by_owner` loads itself lazily, so the singleton readers were always correct — which is why macroscope's own "assigned nil" report on this line was wrong — but only `#call` reaches `Batch`, which is what *runs* each loader, and running is what writes into the association cache. 6.1's `#preload` always did both. So the two versions disagreed on what a preloaded polymorphic relation leaves behind, and `becomes()` dropped the target on 7+ only.

`#call` now, on 7+. No extra query — `records_by_owner` was issuing it one line later anyway; the existing "exactly 1 query, no JOIN" spec still holds on all six versions. Pinned by a spec that reads the target through the converted record under `uncached` (the query cache would otherwise swallow the very SELECT that proves the loss); it fails on 7.1 with `#loaders`.

Comment narrowed to say what actually crosses and why, rather than claiming blanket coverage.

app/services/forest_liana/base_getter.rb
christophebrun-forest: Confirmed on all three, and corrected in the description.

Neutralising the owner-key selection, reproduced:

| scenario | 6.1 | 7.1 / 8.1 |
|---|---|---|
| `has_many … primary_key: 'name'` | `MissingAttributeError: missing attribute: name` | **no error**, `values=[[]]` |
| `has_one … through:` | `missing attribute: island_id` | `missing attribute 'island_id' for Tree` |

So "each on the missing-attribute 500" held only on 6.1, exactly as you said. The Safety section now carries the version split, with the mechanism (`derive_key` reads the key with no raising block, where `BelongsToAssociation#stale_state` passes one) — it makes the argument for selecting the key stronger, not weaker: on modern Rails the same author mistake degrades to silently wrong data instead of to a visible failure.

Also noted there that the specs assert on values and query counts, so they catch it on every version regardless.

app/services/forest_liana/base_getter.rb
```diff
@@ -41,6 +41,95 @@ def optimize_record_loading(resource, records, force_preload = true)
...
+    # Preloader#preload takes the same records/associations positionally and returns the loaders
+    # directly (one per target class among the polymorphic records), without that branch grouping
+    # - preloaded one association at a time here so its name is already known, not read back off
+    # a branch this version's Preloader doesn't expose.
+    def preload_polymorphic_associations(records, associations)
+      return if associations.empty? || records.empty?
+
+      if Rails::VERSION::MAJOR >= 7
+        preloader = ActiveRecord::Associations::Preloader.new(records: records, associations: associations)
+        preloader.loaders
```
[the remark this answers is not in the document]
christophebrun-forest: Not reproducible — the targets resolve correctly on Rails 7.1:

```
targets=["#<User id: 1 …>", "#<Island id: 1 …>"]
```

`Preloader::Association#records_by_owner` calls `load_records` itself when the loader has not run in a batch, so reading it off the loader is what performs the preload; nothing is assigned `nil`.

You were right that `#call` was missing, though, for a different reason: only `#call` reaches `Batch`, which *runs* each loader, and running is what writes the target into the association cache. Without it Rails 7+ left nothing for `becomes()` to carry in `get_record`, where 6.1's `#preload` did. Changed to `#call` in 2d37749 for that, and pinned by a spec in `sti_record_conversion_spec.rb` — see the thread on `resources_controller.rb`.

matthv:
`base_getter.rb:115-118` — this justification is wrong:

```ruby
# ActiveRecordError covers HasManyThroughAssociationNotFoundError and its siblings, which
# #klass raises for a chain naming a hop that no longer exists...
```

Reproduced against actual activerecord source (6.1.7.9, 7.0.10): a broken `:through` chain makes `#klass` raise `NoMethodError` (already caught by `rescue NameError`), not `HasManyThroughAssociationNotFoundError`. That error only comes from `check_validity!`, called via `record.association(name)` — never invoked here. No spec exercises this branch either.

Keeping `ActiveRecordError` is fine as a defensive catch, but the comment names an exception this method can't actually receive. Worth rewording, or dropping the claim.

matthv (approved):

christophebrun-forest:
Confirmed and corrected in f45299f — the comment was wrong, on all three versions I could check.

Reproduced against 6.1.7.9, 7.0.10 and 8.1.3.1, on a `has_many … through:` naming a hop that does not exist and on one whose `source:` does not:

| shape | `#klass` raises | `NameError`? | `ActiveRecordError`? |
|---|---|---|---|
| missing through hop | `NoMethodError: undefined method `klass' for nil` | ✅ | ❌ |
| missing source | `NoMethodError: undefined method `class_name' for nil` | ✅ | ❌ |
| `class_name` pointing at no model | `NoMethodError` / `NameError` (message varies by version) | ✅ | ❌ |

`HasManyThroughAssociationNotFoundError` only comes from `check_validity!`, and nothing this method reads off a reflection calls it — exactly as you said. So `ActiveRecordError` catches nothing known, and naming that error as its justification was backwards.

Reworded to say what is true: every reachable shape arrives as a `NameError`, `ActiveRecordError` is kept as a net rather than for a caller, and whatever does land there is named in the log line instead of being silently indistinguishable from "the relation does not exist". Kept rather than dropped, per your "fine as a defensive catch".

You were also right that nothing exercised the branch. It has a spec now, stubbing the `NoMethodError` the repro above produces and asserting the path degrades *and* logs; it fails if the rescue stops matching. Stubbed rather than declared on a real broken model, which would have to live in the dummy app and be counted by every spec that walks `ActiveRecord::Base.descendants`.

881 examples pass on all six versions.

matthv (approved):

PR #806, opened 2026-09-15
test(search): cover acts_as_taggable_on search on every CI leg, and stop re-running its DSL per request

matthv:
## Context

PR #797 rewrote `SearchQueryBuilder#acts_as_taggable_query` (an executed id list → a deferred `.to_sql` subquery) and fixed a bind-scanning crash on a tag name containing a colon. None of it was exercised against a real `acts_as_taggable_on` relation: the gem wasn't in any of the 6 CI gemfiles, and the one spec covering `acts_as_taggable_query` called it directly against a hand-built `Tree.where(...)` relation standing in for `tagged_records`, not a real `tagged_with(...)` result.

## What changed

- `acts_as_taggable_on` installed on every CI leg — pinned per gemfile since no single major covers Rails 6.1 through 8.1 (10.x for 6.1-7.1, 12.x for 7.2-8.0, 13.x for 8.1, the only major reaching 8.1). Its own `Tag`/`Tagging` models are excluded from the dummy app's apimap the same way any other internal join model would be.
- A real taggable model in the dummy app (`Article`, single-context) — not `Tree`, which is what corrupted `ActiveRecord::Delegation`'s per-class method cache the last time this was tried directly on it (#797).
- `search_query_builder.rb` and `resource_deserializer.rb` were both calling `acts_as_taggable` (no args) on every request — which re-runs the gem's whole `taggable_on` macro (re-including modules, redefining methods on the model) each time. Both now read `tag_types`, the same information already persisted as a class_attribute by the gem's own bootstrap-time call. Also drops `acts_as_taggable?(field)`, which had no caller.
- `search_query_builder_spec.rb`'s `#acts_as_taggable_query` unit test (exercised against the `Tree.where(...)` stand-in) is replaced by an end-to-end `SearchQueryBuilder#perform` suite against `Article`: a term matching a tag via a single subquery (not an executed id list), a term matching nothing (emptied, not the whole table), the colon-in-a-tag-name bind-scanning regression #797 fixed, a guard that the gem's macro never re-runs, and a non-taggable resource left unaffected.

## Verification

806 examples, 0 failures, green on every supported Rails version (6.1 through 8.1).

fixes PRD-1190

> [!NOTE]
> ### Stop re-running `acts_as_taggable` DSL per request in `SearchQueryBuilder` and `ResourceDeserializer`
> - `ResourceDeserializer#acts_as_taggable_attribute?` now reads `tag_types` instead of invoking the `acts_as_taggable` macro to detect tag contexts.
> - `SearchQueryBuilder#search_param` now guards on `taggable?`, calls `tagged_with` once, and builds a single tag condition across all contexts instead of looping per field and checking `respond_to?(:acts_as_taggable)`.
> - Adds `acts-as-taggable-on` to the test group across every Rails gemfile (10.x–13.x) and introduces an `Article` dummy model with a migration for `articles`, `tags`, and `taggings` tables.
> - Replaces old `Tree`-based search specs with end-to-end `Article` examples covering tag matching, unknown tags, tags containing colons, and confirmation the macro is not invoked during search.
> - Risk: `SearchQueryBuilder` no longer checks `respond_to?(:acts_as_taggable)`, so models that are taggable but do not respond to `tagged_with` will raise instead of silently skipping the tag-search branch.
>
> 
> #### Changes since #806 opened
>
> - Added `respond_to?(:tagged_with)` guard condition to `ForestLiana::SearchQueryBuilder.perform` method's ActsAsTaggable branch [85309cc]
> - Added SQL query assertion to ActsAsTaggable search test case [85309cc]
> - Changed ActsAsTaggable attribute detection and iteration in `ForestLiana::SerializerFactory` and `ForestLiana::SchemaAdapter.perform` [9fbde78]
> - Added test coverage for `ForestLiana::SearchQueryBuilder.acts_as_taggable_query` [9fbde78]
> - Updated comments in `ForestLiana::SearchQueryBuilder.acts_as_taggable_query` [9fbde78]
> - Changed the `acts_as_taggable_on` gem detection in `ForestLiana::SearchQueryBuilder.perform` from checking `respond_to?(:tagged_with)` to `respond_to?(:tag_types)`, and updated test coverage to verify that models defining custom `taggable?` and `tagged_with` methods without the gem no longer trigger the gem-specific tagged_with call [b2591a8]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 85309cc. 1 file reviewed, 1 issue evaluated, 0 issues filtered, 1 comment posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>


app/services/forest_liana/search_query_builder.rb
```diff
@@ -169,11 +169,13 @@ def search_param
...
-        if @resource.try(:taggable?) && @resource.respond_to?(:acts_as_taggable)
-          @resource.acts_as_taggable.each do |field|
-            tagged_records = @records.tagged_with(@search.downcase)
-            push_condition(tag_conditions, acts_as_taggable_query(tagged_records), @resource.primary_key.to_s)
-          end
+        # taggable? alone is enough of a guard: acts_as_taggable_on redefines it to true on the
+        # first `acts_as_taggable_on` call, so respond_to?(:acts_as_taggable) is already true for
+        # every model once the gem is loaded, taggable or not. tagged_with with no context searches
+        # every context the model declared (tag_types), so this needs calling only once.
+        if @resource.try(:taggable?)
```
christophebrun-forest, line 176: **Blocking.** Dropping the `respond_to?(:acts_as_taggable)` half of the guard widens it in a way the comment doesn't cover: the gem redefines `taggable?` only when it is loaded, but a model can define its own `self.taggable?` for its own reasons while the gem is absent. In that case this branch now fires and `@records.tagged_with` raises `NoMethodError` — a 500 on every search of that collection, where it used to be skipped.

`respond_to?(:tagged_with)` keeps the fix (one call, no macro re-run, no `acts_as_taggable` invocation) without the exposure:

```ruby
if @resource.try(:taggable?) && @resource.respond_to?(:tagged_with)
```

Worth noting the two untouched sites still use the old belt-and-braces form — `app/services/forest_liana/schema_adapter.rb:14` and `app/serializers/forest_liana/serializer_factory.rb:362` — so the codebase ends up inconsistent either way; aligning all three would be better than leaving one narrowed guard among two wide ones.

spec/dummy/config/initializers/forest_liana.rb
```diff
@@ -2,3 +2,7 @@
 ForestLiana.auth_secret = 'auth_secret_test'
 ForestLiana.application_url = 'http://localhost:3000'
 ForestLiana.workflow_executor_url = 'http://workflow-executor.test:4001'
+# acts_as_taggable_on's own tables, not Forest collections — keep them out of the apimap and out
+# of ForestLiana.models the same way any other internal join model would be. ForestLiana.name_for
+# replaces '::' with '__', so the namespaced class name has to be pre-mangled to match here too.
+ForestLiana.excluded_models = %w[ActsAsTaggableOn__Tag ActsAsTaggableOn__Tagging]
```
christophebrun-forest, line 8: Non-blocking. Making `ForestLiana.excluded_models` a load-bearing global default is a bit fragile: `spec/requests/associations_spec.rb:522` correctly saves and restores it, but any spec that resets it to `[]` instead silently un-excludes the gem's models and breaks `bootstrapper_spec`'s exact `match_array` under random order.

That is not hypothetical — `feature/prd-1089` has exactly such a spec (`query_footprint_spec.rb`, `after { ForestLiana.excluded_models = [] }`), so this becomes a real order-dependent flake once both branches land. I'll fix the save/restore on that branch; flagging it here so whoever merges second knows why.

spec/services/forest_liana/search_query_builder_spec.rb
```diff
@@ -15,39 +15,66 @@ module ForestLiana
...
+      it 'matches a tagged record via a single subquery, not an executed id list' do
+        records = nil
+        queries = capture_queries { records = builder.perform(Article.all).to_a }
+
+        expect(records).to eq([tagged])
+        expect(queries.grep(/\ASELECT/).size).to eq(1)
+        expect(queries.first).to match(/articles\.id IN \(SELECT "articles"\."id" FROM "articles"/)
+      end
+
+      it "empties the result on a term matching no tag, rather than fall through to the whole table" do
```
christophebrun-forest, line 39: Non-blocking. This example doesn't isolate the tag path: `'nothing-tagged-this'` also fails the `title LIKE` condition on both articles, so an empty result proves nothing about the tag branch — it would pass identically if the tag subquery were never built at all.

Asserting on the SQL as well as the rows would make it load-bearing, e.g.:

```ruby
expect(records.to_sql).to match(/articles\.id IN \(SELECT/)
expect(records.to_a).to be_empty
```

so the example fails if the branch is skipped, rather than only if it wrongly falls through to the whole table.

christophebrun-forest (commented):
Reviewed the full diff plus the surrounding code, the acts-as-taggable-on sources (v10.0.0 and v13.0.0), the CI matrix and the dummy app's boot/migration path.

One blocking finding and two non-blocking notes, all left inline.

What I checked and cleared:

- `acts_as_taggable` really does return `tag_types.map(&:to_s)` (last expression of the `Core::ClassMethods#taggable_on` override), not the class — so the deserializer's `tag_types.map(&:to_s)` is string-for-string equivalent to the old `acts_as_taggable.to_a`, minus the macro re-run.
- A single `tagged_with` with no `:on` applies no context filter, so it covers every declared context; the old loop just pushed N identical conditions. Same results.
- `tagged_with` returns `none` (`WHERE 1=0`) on an empty parsed tag list, so `reselect(pk).to_sql` stays valid SQL, and `AllTagsQuery` emits no `select`/`group` without `match_all`, so `reselect` is safe.
- The migration is a faithful consolidation of the gem's templates 1/2/4/6/7 (the retained `taggings_taggable_context_idx` is a harmless duplicate), timestamp ordering and `Migration[6.0]` match the rest of the dummy app, and `schema.rb` matches the migration.
- Gem pins line up with each leg's activerecord requirement (10.x for 6.1/7.0/7.1, 12.x for 7.2/8.0, 13.x for 8.1), and declaring them per-gemfile rather than in `common.gemfile` follows that file's stated convention.
- `excluded_models` is set in an app initializer, which runs before the engine's `after_initialize` bootstrapper, `ActsAsTaggableOn__Tag`/`__Tagging` is the right mangling for `name_for`, and `Zeitwerk::Loader.eager_load_all` does make those models AR descendants — so the exclusion is genuinely needed.
- The removed `acts_as_taggable?(field)` has no callers left on the branch.

christophebrun-forest (approved):
Ok, for me.

PR #807, opened 2026-09-16
fix(charts): build a leaderboard from the parent collection so its table reaches the FROM clause

ludovicl:
## Problem

A Leaderboard chart returns a 500 unless the related model happens to `belongs_to` the parent collection.

`LeaderboardStatGetter#perform` queries the related model, but groups and aggregates on the parent table:

```ruby
@scoped_child_model
  .joins(includes)                                        # one-associations of the child only
  .where({ @scoped_parent_model.name.downcase.to_sym => @scoped_parent_model })
  .group(@group_by)                                       # "organisations.customer_success_manager"
  .pluck(@group_by, Arel.sql("#{aggregation_sql(...)}"))  # COUNT(DISTINCT organisations.id)
```

The parent table reaches the FROM clause only through `joins(includes)`, that is only when the child model declares a `belongs_to` on the parent. On our app, on a chart grouping the `Organisation` collection by a string column:

```
PG::UndefinedTable: ERROR: missing FROM-clause entry for table "organisations"
LINE 1: SELECT "organisations"."customer_success_manager", COUNT(DIS...
```

I ran a leaderboard on each of the 66 relationships of that collection: 35 raise this error, including every `has_many :through`. Another 21 raise `ActiveRecord::EagerLoadPolymorphicError`, because `joins(includes)` eager-loads every one-association of the child and some of them are polymorphic. 10 answer.

Two more problems show on the 10 that answer:

- `aggregation_sql` resolves the aggregate field against `@resource`, the parent. A leaderboard summing a field of the related records builds `SUM(organisations.amount_cents)` and raises `PG::UndefinedColumn`. This one is a regression: up to v9.14.4 the aggregate ran on the child relation through `.send(@aggregate, @aggregate_field)`, and #741 replaced it with `aggregation_sql`.
- with the default `count`, it builds `COUNT(DISTINCT organisations.id)`, so the chart shows a number of parent rows where the user asked for a number of related records.

## Fix

Start the query from the parent collection, join the relationship, and aggregate on the related model:

```ruby
@scoped_parent_model
  .joins(@relationship_field_name)
  .merge(@scoped_child_model)
  .group(@group_by)
  .order(Arel.sql("#{alias_name} DESC"))
  .limit(@limit)
  .pluck(@group_by, Arel.sql("#{aggregation} AS #{alias_name}"))
```

`merge` keeps the scope of the related collection, which the previous `where` carried. `resolve_field_path` and `aggregation_sql` take the resource as an optional argument, defaulting to `@resource`, so `PieStatGetter` is untouched.

## Test

`spec/services/forest_liana/leaderboard_stat_getter_spec.rb` is new, the getter had no spec. It covers a relationship the child does not belong to (`User has_many :addresses, as: :addressable`), the counting semantics, and a sum on a field of the related records. The three examples fail on `main`: the dummy app raises `EagerLoadPolymorphicError` and `ArgumentError` where a schema with `has_many :through` relationships raises `PG::UndefinedTable`.

Green on every CI leg, run locally on Ruby 3.3: Rails 6.1, 7.0, 7.1, 7.2, 8.0 and 8.1, each
with `rake db:migrate test` at 23 runs and 28 assertions, and `rspec` at 515 examples, 0 failures.

## Notes

Unrelated to this change, and left alone: `get_scoped_model` passes `@params` to `FiltersParser`, but `initialize` never assigns it, so a scope on a leaderboard collection is parsed without the request context variables.

### General

### Security

`@group_by` and the aggregate field keep going through `resolve_field_path`, which builds them from
the model table name and the reflection, and the aggregator is still restricted to `sum` and
`count`. `merge` carries the scope of the related collection that the previous `where` carried, so a
leaderboard reads no more rows than before.

> [!NOTE]
> ### Fix leaderboard charts by querying from parent collection so its table reaches the FROM clause
> - `LeaderboardStatGetter#perform` now starts from the scoped parent model, joins the configured relationship, merges the child-model scope, and builds aggregate SQL against the child model — replacing the previous child-rooted query that could not reach certain parent-side relationships
> - `AggregationHelper#resolve_field_path` and `#aggregation_sql` accept an optional `resource` argument so aggregate expressions can target a resource other than the helper's instance resource; callers that omit it retain existing behavior
> - Adds specs covering scope-cache setup and leaderboard relationships in both directions (user→address and island→tree)
> - Behavioral Change: leaderboard queries now group parent records while counting or summing child fields; any callers relying on the old child-rooted query shape in `leaderboard_stat_getter.rb` will see different SQL
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 249a430.</sup>
> 
>


matthv:
Hi @ludovicl, thanks for this fix 💪 

We currently have WIP in the same area, so we'll hold off reviewing in depth for a few days to avoid conflicting feedback.
We'll come back to it then. 

Thanks for your patience!

matthv:
Thanks, the diagnosis and the direction are right.

Testing against a scoped collection surfaced three problems. The spec's `scopes => {}` fixture never exercises one, which is why they pass unnoticed today.

**1. `.merge` re-roots the child scope's joins onto the parent.** A scope filtering a related field (`owner:name`) lands in the child's `eager_load_values`; `merge` copies those onto the parent, where Rails re-resolves them. Either it raises (`Can't join 'Island' to association named 'owner'`), or, when the parent has an association of the same name, it silently re-points:

```sql
FROM "isle" INNER JOIN "trees" ON "trees"."island_id" = "isle"."id"
            LEFT OUTER JOIN "users" ON "users"."id" = "isle"."id"   -- the PARENT's owner
WHERE (users."name" = 'Aegon')
```

The scope stops restricting the child: `[Dragonstone 2]` instead of `[Skagos 1]`, rows the viewer is scoped out of.

**2. The parent's `default_scope` ordering survives, and `.order` appends instead of replacing**, giving `ORDER BY hired_at ASC, count_id DESC`. On PostgreSQL that is a hard failure, not just a mis-ranking, since `hired_at` is not in the `GROUP BY`. CI is SQLite-only, so it passes here.

**3. `.merge` is pinned by nothing**: deleting it leaves the suite green at 934/0.

A direction that works, verified on Rails 6.1 through 8.1 with public API only:

```ruby
if @child_scope
  key = @child_model.primary_key
  records = records.where(@child_model.table_name => { key => @child_scope.select(key) })
end

records.group(@group_by).reorder(Arel.sql("#{alias_name} DESC"))
```

An `INNER JOIN` variant looks equivalent but drops rows for `blank` and `not_equal`; the subquery keeps the `LEFT OUTER` semantics `FiltersParser` emits.

Worth a spec: a scoped child collection, and a scoped parent whose `default_scope` orders. The dummy's `Owner` already has one.

PR #808, opened 2026-09-21
feat: add skip rbac options

arnaud-moncel:
### General

### Security

> [!NOTE]
> ### Add skip option for relation read-permission RBAC checks
> - Adds a module-level `skip_relation_read_permission` configuration option to `ForestLiana`, defaulting to `false`.
> - When enabled, `ForestLiana::Ability::Permission#read_permissions` bypasses the permission-system fetch and treats every requested collection as readable.
> - Updates `ForestLiana::CapabilitiesGetter` to report the relation-read-permission status in the agent capabilities response, and adds a boot-time warning when the option is active.
> - Risk: Disabling relation read-permission checks exposes columns reached through relation paths without verifying read permissions on the target collection. Collection-level CRUD authorization remains active.
>
> 
> #### Changes since #808 opened
>
> - Added strict boolean predicate for RBAC skip option and changed permission caching behavior [0b2a546]
> - Changed capabilities payload to use strict predicate for relation read permission checking flag [0b2a546]
> - Changed security warning log to use strict predicate [0b2a546]
> - Added test coverage for strict boolean handling and apimap exposure checks [0b2a546]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7538466.</sup>
> 
>


app/services/forest_liana/ability/permission.rb
```diff
@@ -60,7 +60,10 @@ def read_permissions(user, collection_names)
           # An absent permission system allows everything, so it is not queried: `is_crud_authorized?`
           # short-circuits the same way, and answering anything else here would redact every relation
           # on a deployment that granted nothing to check.
-          if has_permission_system?
+          # An absent permission system and `skip_relation_read_permissions` ask for the same
+          # answer — the first because nothing was granted to check, the second because the
+          # operator turned the checks off on purpose.
+          if has_permission_system? && !ForestLiana.skip_relation_read_permissions
```
christophebrun-forest: The conditions should be reversed to avoid calling the endpoint to evaluate 'has_permission_system?'

christophebrun-forest: **The fetch still happens when the option is on.**

`&&` evaluates left to right, so `has_permission_system?` runs first. That method hits the `forest.has_permission` cache, and on a cold cache it calls `/liana/v4/permissions/environment` — which can raise `HTTP403Error`. So with `skip_relation_read_permissions = true`, the very network round-trip the option is meant to remove still fires on the first request after boot (or after TTL expiry).

This contradicts both the PR description ("bypasses the permission-system fetch") and the spec title "without fetching permissions" — that spec pre-warms the cache, so it never exercises the cold path.

agent-ruby v2 has the operands the other way round in the equivalent code (`services/permissions.rb:70`):

```ruby
if skip_relation_read_permissions? || !permission_system?
```

Suggest mirroring it:

```ruby
if !ForestLiana.skip_relation_read_permissions && has_permission_system?
```

christophebrun-forest: **Separate concern: this also disables apimap-exclusion enforcement, which isn't an RBAC read permission.**

When this branch is skipped, the `else` marks every requested collection readable in `@read_permissions_cache`. That makes the `UnexposedQueryCollectionError` / `denial_entry(unexposed:)` branches further down (around L166) unreachable.

Concrete effect: a filter, sort or search traversing a relation to a collection removed from the apimap via `ForestLiana.excluded_models` used to be rejected; it now runs against that table. `FiltersParser` resolves relation paths through `reflect_on_association`, which knows nothing about the apimap, so there is no second line of defence.

Exclusion is an exposure decision, not a role permission — the option shouldn't switch it off. Worth keeping the unexposed check outside the skipped branch.

lib/forest_liana.rb
```diff
@@ -30,6 +30,7 @@ module UserSpace
   mattr_accessor :logger
   mattr_accessor :reporter
   mattr_accessor :skip_schema_update
+  mattr_accessor :skip_relation_read_permissions
```
christophebrun-forest, line 33: **This flag is read by truthiness, unlike every comparable one.**

The three consumers (`permission.rb:66`, `capabilities_getter.rb:78`, `bootstrapper.rb:31`) all test the raw value. Its neighbour one line above is compared strictly — `bootstrapper.rb:697`:

```ruby
ForestLiana.skip_schema_update == true
```

and agent-ruby v2 does the same at all four of its read sites (`permissions.rb:280`, `collections.rb:89`, `charts.rb:59`, `agent_factory.rb:349`):

```ruby
config[:skip_relation_read_permissions] == true
```

The generated initializer is plain Ruby and the ENV pattern is already the documented one for secrets, so this is a realistic line for an operator to write:

```ruby
ForestLiana.skip_relation_read_permissions = ENV['FOREST_SKIP_RELATION_READ_PERMISSIONS']
```

`"false"`, `"0"` and `""` are all truthy in Ruby, so setting the variable to `false` — or leaving it defined but empty in a `.env` — silently *disables* the checks. It fails open, and it's silent on the front too: the capability goes out as `checksRelationReadPermissions: false`, so Ember stops pruning projections as well.

Suggest a single predicate, consistent with the other two SDKs:

```ruby
def self.skip_relation_read_permissions?
  skip_relation_read_permissions == true
end
```

and using it at the three call sites. It also gives a natural place to fix the operand order flagged in `permission.rb`.

app/services/forest_liana/capabilities_getter.rb
```diff
@@ -64,13 +64,21 @@ def initialize(collection_names)
...
+        agentCapabilities: agent_capabilities,
         collections: requested_collections.map { |collection| collection_capabilities(collection) }
       }
     end
 
     private
 
+    # The frontend prunes projections by the role's read permission on this flag, so it follows the
+    # option rather than sitting in the frozen constant beside the static ones.
+    def agent_capabilities
```
christophebrun-forest, line 77: **The class NOTICE above now contradicts the spec NOTICE updated in this same PR.**

`capabilities_spec.rb` is changed here to say `canUseProjectionViaHeaderOnList` "no longer doubles as the pruning floor: the frontend reads checksRelationReadPermissions for that, and this liana does implement the pruning."

But the NOTICE at the top of this file (L6-11) still states the opposite — that the flag *is* the pruning floor and that the policy is unimplemented (PRD-1083):

> canUseProjectionViaHeaderOnList is held back although the routes do honour the header on the list: the frontend also reads it as a version floor for pruning projections by the role's read permission, a policy this agent does not implement (PRD-1083).

Since that NOTICE is the stated reason the flag stays `false`, leaving it stale means the next reader can't tell why it's still held back. Worth updating it in the same pass.

christophebrun-forest (commented):
Reviewed the new `skip_relation_read_permissions` option. The contract name and the leaf-collection semantics line up with the Ember front (`capabilities.ts`, `reference-permissions.ts`) and with the agent-ruby v2 equivalent, and the 8 new examples pass (the other failures in these 4 spec files reproduce identically on the base commit, so nothing regressed).

Four points below — the first two are behavioural, the last two are consistency.

christophebrun-forest (approved):
Look Good To Me

PR #812, opened 2026-09-23
fix(apimap-sorter): keep sorting when fields share a name with different types

ramongtx:
## Problem

`ApimapSorter#sort_array_of_fields` compares `[field, type]` tuples directly:

```ruby
[field1['field'], field1['type']] <=> [field2['field'], field2['type']]
```

A field `type` is a `String` (`"String"`), an `Array` (`["String"]` for an array column) or a
`Hash` (a nested type). Those are not comparable with each other — `["String"] <=> "String"`
returns `nil` — so as soon as two entries share a `field` name, the sort block returns `nil` and
`Array#sort` raises `ArgumentError: comparison of Hash with Hash failed` (the message names the
sorted elements, which are the field hashes).

`perform` rescues that and returns the apimap, but the failure happens inside `Array#map!`, which
has already been interrupted. The collections it had not reached yet are never passed through
`reorder_collection_attributes`, so they keep attributes outside `KEYS_COLLECTION` — in practice
`search_fields`, which the `search_fields` collection DSL sets and `SchemaFileUpdater` deliberately
excludes from `.forestadmin-schema.json`.

The Forest API then rejects the entire apimap:

```json
{"errors":[{"status":400,"detail":"Validation failed","meta":{"message":
"[{\"instancePath\":\"/data/84/attributes/searchFields\",\"keyword\":\"type\",
  \"params\":{\"type\":\"string\"},\"message\":\"must be string\"}]"},
"name":"ValidationFailedError"}]}
```

Because `Bootstrapper#send_apimap` discards the response body, the only user-visible symptom is:

```
Forest 🌳🌳🌳  An Apimap reordering issue occured: comparison of Hash with Hash failed
Forest 🌳🌳🌳  An error occured with the apimap sent to Forest. Please contact support@forestadmin.com
```

The schema silently stops reaching the environment and the UI shows "This component is not
configured yet". On our app (385 collections, 81 of them using `search_fields`) this reproduced on
every single boot.

## Fix

Compare the JSON representation of the values, which is total for any input:

```ruby
def comparison_key(object, *keys)
  keys.map { |key| object[key].to_json }
end
```

Plain strings keep their previous relative order (the leading quote is common to both operands), so
already-sorted apimaps do not churn. `sort_array_of_objects` gets the same treatment since it has
the identical latent issue on `nil` ids.

## Verification

Reproduced against a real 385-collection apimap. With the patch applied, `search_fields` is stripped
from all 385 collections by `reorder_collection_attributes` as intended, and
`POST /forest/apimaps` returns **HTTP 202 Accepted** where it previously returned **400**.

4 regression specs added to `spec/services/forest_liana/apimap_sorter_spec.rb`. They fail on `main`
and pass with the fix:

```
16 examples, 4 failures   # without the fix
16 examples, 0 failures   # with the fix
```

Full suite: `516 examples, 0 failures` (Ruby 3.3.6).

The specs cover the two behaviours that matter: the offending collection's fields do get sorted, and
— more importantly — collections *after* it are still processed, so no non-schema attribute leaks
into the payload.

## Unrelated suggestion

Consider logging the response body on a 4xx in `Bootstrapper#send_apimap`. The
`Net::HTTPBadRequest` branch currently prints only "please contact support" and throws away a
response that names the offending collection and attribute precisely. Surfacing it would have turned
this investigation into a one-line diagnosis. Happy to open a separate PR if you want it.

> [!NOTE]
> ### Fix `ApimapSorter` crash when fields share a name with different types
> - Fixes an `ArgumentError` when sorting fields that share the same name but have different Ruby types (String, Array, or Hash), which are not directly comparable.
> - Adds a `comparison_key` helper in [apimap_sorter.rb](https://github.com/ForestAdmin/forest-rails/pull/812/files#diff-177c72307171d214d6643440839f8baadc48cc38dbb7d59ae5b19714f18ebd6b) that converts each value to its JSON string representation before building the sort key, giving all values a common comparable type while preserving prior ordering of plain strings.
> - Updates both `sort_array_of_objects` and `sort_array_of_fields` to derive comparison values through the new helper instead of comparing raw objects.
> - Risk: sort order for mixed-type fields is now deterministic but differs from the previous behavior where such comparisons raised an error; any downstream code relying on the exception will no longer receive it.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e564b3f.</sup>
> 
>


christophebrun-forest (commented):
Thanks for the writeup — the diagnosis is spot on, and the repro is unusually clear.

I checked out the branch, merged `main` into it and ran everything: the 4 new specs fail on `main` (`comparison of Hash with Hash failed`) and pass with the fix, and the full suite is green (**935 examples, 0 failures**). No conflicts.

Two things before merging, both inline. Heads-up too that CI hasn't actually run here — fork PR, the workflows need a maintainer approval; we'll trigger them.

app/services/forest_liana/apimap_sorter.rb
```diff
@@ -111,16 +111,25 @@ def perform
...
     end
 
+    # NOTICE: A field 'type' is a String ('String'), an Array (['String'] for an array
+    # column) or a Hash (nested type), and those are not comparable with one another:
+    # `['String'] <=> 'String'` returns nil, which makes `sort` raise an ArgumentError.
+    # Comparing the JSON representation keeps the ordering total whatever the values are.
+    # Plain strings keep their previous relative order, so sorted apimaps do not churn.
+    def comparison_key(object, *keys)
+      keys.map { |key| object[key].to_json }
+    end
```
christophebrun-forest: > Plain strings keep their previous relative order, so sorted apimaps do not churn.

This one doesn't hold. The leading quote is common to both operands, but the **trailing** one isn't, and JSON escaping changes bytes too:

```
raw     : ["Full", "Full name", "a", "a ", "a!", "a\"b", "ab"]
to_json : ["Full name", "Full", "a ", "a!", "a", "a\"b", "ab"]
```

Any name that is a strict prefix of another followed by a space or `!` flips — plausible for smart field names (`Full` / `Full name`). The impact is cosmetic churn rather than breakage, but someone will rely on that comment later.

Only the tie-breaker ever needs to be total, so JSON-ifying just that one keeps the primary key byte-for-byte identical to today:

```ruby
def sort_array_of_objects(array)
  array.replace(array.sort_by { |element| [element['type'].to_s, element['id'].to_s] })
end

def sort_array_of_fields(array)
  array.sort_by { |field| [field['field'].to_s, field['type'].to_json] }
end
```

I ran this against your specs: 17 examples, 0 failures, including `should keep a deterministic order between the non-comparable types`. It makes the no-churn claim literally true and drops the helper entirely. As a side effect `sort_by` computes the keys once per element instead of once per comparison — 40ms → 8ms on a 385-collection apimap, so it also removes the O(n log n) `to_json` calls your version adds on every boot.

(`sort!` returns `nil` when the array is already sorted, hence the `replace` for the in-place one.)

If you'd rather keep `comparison_key`, that's fine too — just drop the last line of the comment.

spec/services/forest_liana/apimap_sorter_spec.rb
```diff
@@ -171,6 +171,70 @@ module ForestLiana
...
+        # directly, so `['String'] <=> 'String'` returned nil and `sort` raised
+        # `ArgumentError: comparison of Hash with Hash failed`. The rescue in `perform`
+        # swallowed it and returned a half-processed apimap: collections that were never
+        # reached kept attributes outside KEYS_COLLECTION (such as 'search_fields'), and
+        # the Forest API rejects those with `HTTP 400 ValidationFailedError`.
+        apimap = {
+          'meta': {
+            liana: 'forest-rails',
+            'liana_version': '1.5.24',
+          },
```
christophebrun-forest: The fixture has no `meta.stack`, so `perform` raises `undefined method 'keys' for nil` on `@apimap['meta']['stack'] = reorder_keys_basic(...)` and falls into the `rescue`. Running just this context:

```
Forest 🌳🌳🌳  An Apimap reordering issue occured: undefined method `keys' for nil
4 examples, 0 failures
```

The assertions still pass, because the raise happens after the `data` loop — but a regression test whose whole point is "the sorter no longer bails out mid-way" shouldn't itself be bailing out. Could you add `stack:` to the fixture like the first context does?

And while you're there, asserting the invariant directly would make the spec a lot stronger than the indirect checks:

```ruby
expect(FOREST_LOGGER).not_to receive(:warn)
```

PR #813, opened 2026-09-23
chore: trigger the release of the RBAC capabilities work [force release]

matthv:
## Why

The squash commit of #809 (`b89d6f98`) failed commitlint:

```
feat: RBAC capabilities — permissions, projection, search footprint & smart-field preloading (#809)
✖ subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```

The trigger is the leading acronym, not the em dash or the `&` — reproduced locally against `commitlint.config.js`: moving `RBAC` out of first position passes, keeping it there fails.

`deploy` has `needs: [lint, test]`, so semantic-release never ran. `main` is still on **9.21.0** and the gem was never published.

## Why an empty commit

The message is already in `main`'s history and the branch is protected, so correcting it would mean rewriting a branch others have pulled — disproportionate for a casing rule. A push only lints the commits it carries, so a new commit unblocks the pipeline and leaves `b89d6f98` alone.

semantic-release analyses everything since `v9.21.0`, finds the `feat:` from #809 and cuts **9.22.0**. The `[force release]` marker (see `.releaserc.js`) only forces a patch, which the `feat` outranks — it is there to guarantee a release happens, not to set the number.

## Note for next time

The squash message is prefilled from the **PR title**, so that is what has to satisfy commitlint. An acronym in first position fails every time.

> [!NOTE]
> ### Trigger force release of RBAC capabilities work
> Contains only blank-line modifications to force a new release. No functional changes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e8f7097.</sup>
> 
>


christophebrun-forest (approved):
LGTM

